PR #283, opened 2026-04-19
fix: remove /simple/ suffix from PyPI setup page URLs

### LINKED ISSUE #219: Setup page PyPi repository generates wrong URL if repository uses default upstream URL
### Description

If you configure a remote repository for PyPi and use the default upstream URL of https://pypi.org/simple the setup page instructs the user to set the follow environment variable:

`export PIP_INDEX_URL=https://artifact-keeper.apps.cd01.chp.belastingdienst.nl/pypi/<repo-key>/simple`

This will cause artifact-keeper to query upstream at https://pypi.org/simple/simple which will fail.

Leaving the extra `/simple` of will fix this issue.

### Environment Information

- Browser: (e.g., Chrome, Firefox, Safari, Edge)
- Browser Version:
- Operating System:
- Screen Resolution:
- Connection type: (direct, behind proxy)

### Web UI Version

dev

### Backend Version

dev

### Expected Behavior

Clients succeed is downloading libraries from upstream PyPi

### Observed Behavior

Clients fail to download libraries from PyPi.

### Steps to Reproduce

1. Setup remote PyPi repository and use default upstream URL.
2. export `PIP_INDEX_URL` as suggested
3. Run `pip install`.

### Browser Console Logs

```shell

```

### Screenshots

_No response_

### Additional Context

So either the default upstream needs to be changed to just `https://pypi.org` or all the other logic needs to stop adding `/simple` at the end.

### Pre-submission Checklist

brandonrc:
## Summary

Removes the `/simple/` suffix from the pip index URLs generated on the Setup page. When a PyPI remote repository uses the default upstream URL (`https://pypi.org/simple`), the backend appends the request path to it. With the old `/simple/` suffix in the client URL, pip requests would hit `/simple/simple/<package>/` on the upstream, which fails. The correct client-facing URL is `/pypi/{repo_key}/` without the suffix, since the backend handles the simple API path routing internally.

Fixes #219

## Test Checklist
- [x] Manually tested locally

## UI Changes
- [ ] Dark mode verified
- [x] N/A - no UI changes

### DIFF
--- a/src/app/(app)/setup/page.tsx
+++ b/src/app/(app)/setup/page.tsx
@@ -98,12 +98,12 @@ npm config set //${REGISTRY_HOST}/npm/${repoKey}/:_authToken YOUR_TOKEN`,
           title: "Configure pip",
           description: "Add to ~/.pip/pip.conf or ~/.config/pip/pip.conf:",
           code: `[global]
-index-url = ${REGISTRY_URL}/pypi/${repoKey}/simple/
+index-url = ${REGISTRY_URL}/pypi/${repoKey}/
 trusted-host = ${REGISTRY_HOST}`,
         },
         {
           title: "Install a package",
-          code: `pip install --index-url ${REGISTRY_URL}/pypi/${repoKey}/simple/ <package-name>`,
+          code: `pip install --index-url ${REGISTRY_URL}/pypi/${repoKey}/ <package-name>`,
         },
         {
           title: "Upload with twine",


brandonrc:
Closing this PR. The /simple/ suffix is required by PEP 503. pip appends the package name to the index-url, so removing /simple/ causes pip to resolve to /pypi/{key}/{package}/ which doesn't match the backend routes. The original code was correct. The actual issue #219 may be about a different URL generation problem that needs further investigation.

PR #335, opened 2026-05-04
fix(settings): wire Storage tab to runtime config

### LINKED ISSUE #334: Admin Settings page shows hardcoded placeholder values, not runtime config
## Summary

The admin Settings page (`/settings`) displays a banner stating *"Server settings are configured via environment variables. The values shown below reflect the current runtime configuration."* — but several fields are hardcoded literals in the React component, not pulled from `/api/v1/admin/settings`. The displayed values are therefore wrong on any deployment that doesn't happen to match the placeholders.

## Affected file

`src/app/(app)/(admin)/settings/page.tsx`

## Verified on

- `main` at `304cb71` (current HEAD as of filing).
- Published image `ghcr.io/artifact-keeper/artifact-keeper-web:latest`, which resolves to `sha256:33d9d4d245a6…` — the same digest as tags `1.1.3` and `1.1`.

## Reproduction

1. Deploy AK with `STORAGE_BACKEND=s3`, `S3_BUCKET=…`, `STORAGE_PATH=/data/storage` (any non-default path).
2. Confirm the backend honors the config:
   - Backend logs include `S3 storage backend initialized`.
   - `GET /api/v1/admin/settings` returns `{"storage_backend":"s3","storage_path":"/data/storage", …}`.
3. Open the admin Settings → Storage tab in the web UI.

## Expected

UI displays `Storage Backend: S3` and `Storage Path: /data/storage` (matching the API response).

## Actual

UI displays `Storage Backend: Local Filesystem` and `Storage Path: /data/artifacts`, regardless of the actual configuration.

## Affected fields

Hardcoded `SettingRow` values in `settings/page.tsx`:

| Tab | Label | Hardcoded value | Source of truth |
|---|---|---|---|
| Storage | Storage Backend | `Local Filesystem` | `state.config.storage_backend` (returned by `/api/v1/admin/settings`) |
| Storage | Storage Path | `/data/artifacts` | `state.config.storage_path` (returned by same endpoint) |
| Storage | Max Upload Size | `5 GB` | `max_upload_size_bytes` (returned by same endpoint) |
| Storage | Deduplication | `Enabled (SHA-256)` | Currently not in API response |
| Auth | Authentication Method | `JWT (JSON Web Token)` | Always true today; harmless |
| Auth | Access Token Expiry | `1 hour` | `jwt_access_token_expiry_minutes` (default 30) — **not currently in API response** |
| Auth | Refresh Token Expiry | `7 days` | `jwt_refresh_token_expiry_days` — **not currently in API response** |
| General | Environment | `Production` (Badge) | No clear source — possibly `process.env.NODE_ENV` |

## Root cause

The page was authored with placeholder values and never wired to a runtime data source. The Password Policy and SMTP fields on the same page are correctly wired through `settingsApi.getPasswordPolicy()` / `settingsApi.getSmtpConfig()` — both of which call the same `/api/v1/admin/settings` endpoint via the SDK's `getSettings()`. The Storage / Auth Token Expiry / Environment fields just don't follow the same pattern.

## Proposed fix (frontend-only fields)

For the fields whose source of truth is already in the API response (Storage Backend, Storage Path, Max Upload Size), wire them via the same `getSettings()` SDK call already used elsewhere in `src/lib/api/settings.ts`. Format the values for display (e.g., `"s3"` → `"S3"`, byte count → human-readable size).

I'm happy to submit a PR scoped to the Storage tab as a first step. The fix is small and isolated, so this might be a candidate for `good first issue` if a maintainer wants to keep it open for a community contributor instead.

## Out of scope for an immediate frontend fix

- **Auth Token Expiry fields.** Fixing these requires the backend to expose `jwt_access_token_expiry_minutes` and `jwt_refresh_token_expiry_days` on `/api/v1/admin/settings`. Should be a separate issue against the backend repo.
- **Deduplication.** Same — needs a backend field.
- **Environment badge.** Could be hardcoded to "—" or removed; the concept doesn't map to a server-side setting.

## Environment

- `artifact-keeper-web`: `latest` / `1.1.3` (digest `sha256:33d9d4d245a6…`)
- `artifact-keeper-backend`: `1.1-dev`
- Deployed via Helm to AWS EKS

ssorathia:
## Summary

Partial fix for #334 — wires the admin Settings → Storage tab to the runtime config returned by `/api/v1/admin/settings` instead of the hardcoded placeholders. Same raw-cast + type-guard pattern that `getPasswordPolicy` and `getSmtpConfig` already use.

Scope is intentionally narrow: only the three Storage-tab fields whose source of truth already exists in the API response (Storage Backend, Storage Path, Max Upload Size). The other items called out in #334 (Auth token expiry fields, Deduplication, General → Environment badge) need backend changes first and are left out of this PR per the issue's "Out of scope" section.

## Changes

- `src/lib/api/settings.ts` — add `StorageSettings` type, `DEFAULT_STORAGE_SETTINGS`, and `settingsApi.getStorageSettings()`.
- `src/app/(app)/(admin)/settings/page.tsx` — wire the Storage tab's first three `SettingRow`s to the new query; add `formatStorageBackend` (filesystem/s3/gcs/azure → friendly label, unknown values render as-is) and `formatBytes` helpers.
- Leave the Deduplication row hardcoded with a comment, since the backend doesn't yet expose it on the settings endpoint.

## Test plan

- [x] Five new tests for `getStorageSettings`: happy path, missing fields, wrong types, SDK error, SDK throw.
- [x] Four new tests for the rendered Storage tab: populated values, friendly backend labels (incl. `filesystem` → `Local Filesystem`), unknown backend fall-through, loading state.
- [x] Existing SMTP "populates from loaded config" test updated (queryCallIndex bumped from 3 to 4 since storage-settings now slots in between password-policy and smtp-config).
- [x] Lint, typecheck, unit tests, and build all pass locally.

Fixes #334 (partial — Storage tab only).


ssorathia:
Pushed `44d3851` addressing two findings from a self-review:

1. **Reuse the existing `formatBytes`.** The original commit shipped a local `formatBytes` in `page.tsx` that diverged from `@/lib/utils#formatBytes` (already used in 10+ places — repos, staging, builds, etc.), rendering `"1.0 GB"` where the rest of the app renders `"1 GB"`. Switched to the shared helper. Side effect: dropped the undocumented "0 bytes → Unlimited" semantic that the local copy had invented; values now reflect exactly what the backend returns.

2. **Stop silently falling back to placeholder defaults on error.** The original `getStorageSettings` returned `DEFAULT_STORAGE_SETTINGS` (`"filesystem"` / `"/data/artifacts"` / `"5 GB"`) on SDK error or shape mismatch — i.e., the exact placeholder strings #334 calls out as the bug. An admin couldn't distinguish "this is the real config" from "the API broke, you're looking at fiction." `getStorageSettings` now throws on SDK error / missing required fields, and the page distinguishes `isLoading` ("Loading...") from `isError` ("Unavailable") via the React Query state. `DEFAULT_STORAGE_SETTINGS` removed.

Also refactored the page-test `useQuery` mocks from an ordinal `queryCallIndex` to a `queryKey`-keyed dispatch helper (so adding a future query before storage/smtp won't silently break unrelated tests), and added a regression test asserting the placeholder strings explicitly do NOT appear when the storage query errors.

Lint / typecheck / unit tests / build all clean locally.

Issue #359, opened 2026-05-09
Extend #206 type-safety hardening to remaining lib/api modules


brandonrc:
## Background

Surfaced during R2 review of #206. After #206 lands, every `as unknown as T` and `as never` cast in the 15 in-scope `src/lib/api/` modules has been replaced with adapters / narrowing helpers / zod schemas. But the same patterns survive in 12 sibling modules that were out of #206's scope:

- `src/lib/api/sso.ts` (~30 occurrences)
- `src/lib/api/sbom.ts` (~17)
- `src/lib/api/security.ts` (~17)
- `src/lib/api/replication.ts` (~10)
- `src/lib/api/lifecycle.ts`
- `src/lib/api/dependency-track.ts`
- `src/lib/api/telemetry.ts`
- `src/lib/api/webhooks.ts`
- `src/lib/api/analytics.ts`
- `src/lib/api/monitoring.ts`
- `src/lib/api/promotion.ts`
- `src/lib/api/uploads.ts` (1)

Total: ~153 `as never` casts across 12 files.

## Proposed change

Apply the same playbook from #206:

1. For each file, audit every `as never` / `as unknown as T` site.
2. Build adapter functions for each SDK response type.
3. Use `assertData` from `lib/api/fetch.ts` for empty-body guards.
4. Use `narrowEnum` (also in `fetch.ts` after #206) for string-to-union narrowing.
5. Add zod schemas at trust boundaries that need runtime validation.
6. Update tests with realistic SDK fixtures.

## Acceptance criteria

- `grep "as unknown as\|as never" src/lib/api/ | grep -v __tests__` returns zero matches OR every remaining instance has an inline comment explaining why.
- All tests pass.
- New code coverage on changed lines ≥ 80%.

## Scope notes

- This is a sequel to #206, not a replacement. Keep PRs file-by-file or per-pair to keep review tractable — #206 ended up at ~1500 lines across 17 files which was at the upper edge of reviewable.
- Some modules (sso, sbom, security) are large; split into multiple PRs.

## Related

- #206 (the reference PR establishing the pattern)
- `src/lib/api/fetch.ts` — exports `assertData`, `narrowEnum` for reuse

PR #360, opened 2026-05-09
refactor(api): replace double-casts with adapters, narrowing helpers, and zod

### LINKED ISSUE #206: Type safety erosion: 53 double-casts in API layer
## Problem

All 15 files in `src/lib/api/` use `as unknown as T` double-casts (53 total), bypassing TypeScript's type system entirely. If the backend response shape changes, these casts silently produce incorrect data at runtime instead of failing at compile time.

This defeats the purpose of using TypeScript for the API layer.

## Suggested Fix

1. Generate shared types from the OpenAPI spec (the SDK pipeline already exists)
2. Use the generated types directly instead of casting
3. For responses that need validation, add zod schemas at the API boundary
4. Remove all `as unknown as T` casts and fix resulting type errors

## Source

Quality audit, 2026-03-23.

brandonrc:
## Summary

Closes #206. Replaces all `as unknown as T` and `as never` casts in 15 SDK-backed `src/lib/api/*.ts` modules with adapter functions, Set-backed narrowing helpers (with observability warnings on unknown values), zod schemas at trust boundaries, and an `assertData` empty-body guard.

Public `xxxApi.*()` return types are unchanged so consumer code isn't touched. Tests: 1932/1932 pass. `next build` passes.

## Why

Per #206: 53+ double-casts in `src/lib/api/` were bypassing TypeScript's type system. If the backend response shape changed, these casts silently produced incorrect data at runtime instead of failing at compile time.

## Approach

- **Adapters** for shape mapping (`null → undefined`, string → narrow union) per file (13 files)
- **SDK-direct alias** when shapes match exactly (totp.ts, uploads.ts)
- **zod schemas** at the settings trust boundary (`getPasswordPolicy`, `getSmtpConfig`)
- **`assertData(data, ctx)`** in `fetch.ts` rejects `undefined`/`null`/`''` with a contextual error
- **`narrowEnum<T>(value, allowed, fallback, warn?)`** generic in `fetch.ts` consolidates string→union narrowing across 8 modules

## Review process

Three rounds, per the team-of-four / three-round process:

- **R1 build** — SWE agent removed all 53 listed casts (16 commits, one per file). 4 parallel R1 reviewers (Senior SE, Test, Security, DevOps) flagged: migration.ts unfinished (17 sites still used `as never`), builds.ts adaptBuildModule lossy, permissions narrowing should warn, profile silent fallback, assertData should reject empty string. **Fixed in 7 commits** that finished migration.ts properly + addressed all R1 majors.

- **R2 simplifier + gates** — `code-simplifier:code-simplifier` extracted `narrowEnum` helper to `fetch.ts` (used by 8 modules), consolidated `coerceItemsArray`/`coercePaginated` helpers in migration.ts, replaced mutating `delete` with conditional spread, removed redundant wrappers. -78 lines net. Tests still 1932/1932 pass. Senior SE consistency reviewer caught a real **regression**: `profile.ts adaptUser` was dropping `totp_enabled`. **Fixed.** Documentation/Completeness reviewer noted the issue's 'remaining `as never` in non-api modules' should be filed as a follow-up — done as **#359**.

- **R3 adversarial** (4 fresh-eyes reviewers):
  - **Production parity** flagged a **BLOCKER**: `next build` failed with TS2322 in profile.ts:175 (the simplifier's conditional spread regressed against `SdkCreateApiTokenRequest.scopes` being required). **Fixed** by using a `SdkCreateApiTokenRequestPartial` typed local + cast at the SDK boundary.
  - **Integration regressions** found **3 real consumer regressions**:
    1. `groupsApi.update` was overwriting name with `''` when caller passed only description
    2. `searchApi.adaptSearchResult` was dropping `is_quarantined`/`quarantine_until`/`quarantine_reason` (the QuarantineBadge would silently disappear)
    3. `treeApi.adaptTreeNode` was dropping `metadata` (file-tree size/download badges would silently disappear)
    All 3 fixed.
  - **Edge cases** flagged: `adaptBuildModule` setting `created_at: ''` rendered as "Invalid Date" — **fixed** by threading parent `Build.created_at` as fallback. `coercePaginated` synthesized `total: items.length` lies for paginated bare arrays — added warning. `coerceItemsArray` silently swallowed non-array shapes — added warning.
  - **Rollback safety**: ship, grade A. Pure refactor, trivial revert via `git revert -m 1`.

## Out of scope

- 12 sibling `lib/api/` modules (sso, sbom, security, replication, lifecycle, dependency-track, telemetry, webhooks, analytics, monitoring, promotion, uploads-edge-cases) still have ~150 `as never` casts. Filed as **#359** to extend this work.
- 4 surviving `as unknown as` casts in `migration.ts:221-224` for `MigrationReport` types are intentional — SDK declares `Record<string, unknown>` for all four fields; the cast is documented inline.

## Test plan

- [x] `npm test` — 1932/1932 pass
- [x] `npm run lint` — 0 errors (33 pre-existing warnings, none introduced by this PR)
- [x] `npx tsc --noEmit` — 1 pre-existing error in `permissions.test.ts:154` (down from 2 on main; this PR fixed one as a side-effect)
- [x] `npm run build` — succeeds, 45 routes generated
- [x] Coverage on `src/lib/api/`: 94.65% lines / 91.58% branches
- [x] All 4 R1 reviewers + 4 R2 reviewers + 4 R3 reviewers cleared

## Findings worth noting (filed as follow-ups)

- Per-row `console.warn` flood when 1000-item lists hit unknown enum values — could dedupe via per-call `Set` (low priority)
- `groups.update` could omit fields more elegantly (low priority)
- `coercePaginated` total lie when backend returns paginated bare array — currently warns; could be plumbed properly when SDK regen models the wrapper

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 - **`release/1.1.x` maintenance branch + `:1.1-dev` Docker tag rule** (#331) - mirrors `artifact-keeper#890`; pushes to `release/1.1.x` now publish `ghcr.io/artifact-keeper/artifact-keeper-web:1.1-dev` so the v1.1.x release-gate can test a true v1.1.x web/backend pair.
 
+### Changed
+- **Type-safe API layer — replace double-casts with adapters and zod** (#206) - removed all `as unknown as T` and `as never` casts in 15 of `src/lib/api/*.ts` files. Each SDK call now goes through an adapter function or a Set-backed narrowing helper that warns on unknown enum values. `assertData` (new in `fetch.ts`) rejects empty body responses with a contextual error. `settings.ts` uses zod `.safeParse()` at the trust boundary for `getPasswordPolicy`/`getSmtpConfig`. Public `xxxApi` return types unchanged so consumer code is untouched.
+
 ### Fixed
 - **Mutation errors now surface backend details instead of generic placeholders** (#207) - audited every TanStack Query `useMutation` and replaced opaque `onError: () => toast.error("Failed to ...")` callbacks with `toUserMessage(err, fallback)`-driven toasts. 91 callsites across 27 files. Also adds `onError` to 8 previously-silent mutations (security/policies/scans + repo-selector preview), and disambiguates the SSO toggle toasts per provider (OIDC/LDAP/SAML). `toUserMessage` now also reads FastAPI-style `.detail` fields so plugin-install errors (and any other FastAPI-shaped backend error) surface their server-side message.
 
--- a/src/lib/api/__tests__/admin.test.ts
+++ b/src/lib/api/__tests__/admin.test.ts
@@ -22,11 +22,24 @@ describe("adminApi", () => {
   });
 
   it("getStats returns typed AdminStats", async () => {
-    const stats = { total_repos: 5 };
+    const stats = {
+      total_repositories: 5,
+      total_artifacts: 10,
+      total_storage_bytes: 1024,
+      total_users: 3,
+      active_peers: 0,
+      pending_sync_tasks: 0,
+      total_downloads: 0,
+    };
     mockGetSystemStats.mockResolvedValue({ data: stats, error: undefined });
     const { adminApi } = await import("../admin");
     const result = await adminApi.getStats();
-    expect(result).toEqual(stats);
+    expect(result).toEqual({
+      total_repositories: 5,
+      total_artifacts: 10,
+      total_storage_bytes: 1024,
+      total_users: 3,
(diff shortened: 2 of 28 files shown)


bastizol-wq:
Re: “refactor(api): replace double-casts with adapters, narrowing helpers, and zod” — one practical fix is to treat tools like untrusted inputs and validate the contract first.

If you already have OpenAPI/MCP metadata, you can make an allow/warn/block decision deterministically (and log it) before execution.

ATCP is an API-first gate that evaluates a tool/OpenAPI and returns a structured allow/warn/block result.

If you want, I can share the exact first checks we run (host policies + auth expectations + dangerous ops).

PR #421, opened 2026-05-20
fix(repositories): reset create form when dialog reopens

dragonpaw:
## Summary

The Create Repository dialog kept stale form values across submissions: after creating a repo, clicking "+ Create Repository" again pre-filled the previous repo's key, name, format, etc. instead of showing an empty form.

Root cause: `RepoDialogs` stays mounted for the lifetime of the page, so its `useState` values survive open/close cycles. The existing `resetCreateForm()` only ran via `handleCreateClose`, which is wired to Radix Dialog's `onOpenChange`. Radix fires `onOpenChange` for user-initiated closes (Esc, overlay click, Cancel) but not when the parent flips `open` from `true` to `false` programmatically — which is exactly what `RepositoriesContent` does inside the create mutation's `onSuccess`. The close-time reset was bypassed on every successful create, leaving stale state for the next open.

Fix: reset the create form via `useEffect` whenever `createOpen` transitions to `true`, so every open starts clean regardless of how the previous close happened. Smaller and more robust than threading a success signal into the dialog (which would also risk clearing the form if the submit eventually errors).

## Test Checklist
- [x] Manually tested locally

Added a regression test in `repo-dialogs.test.tsx` that exercises the controlled-open path (open → fill → submit → programmatic close → reopen) without going through `onCreateOpenChange`, and asserts the key/name inputs are empty on the second open. Confirmed to fail on `main` without the fix.

Full test suite: 2119/2119 pass. Lint: 0 errors (warnings unchanged from `main`).

## UI Changes
- [ ] Dark mode verified
- [x] N/A - no UI changes

No visual change; behavior-only fix to form-state hygiene.

## Notes for reviewers

The same bug class likely lurks in the Edit dialog (`onOpenChange`-only reset of edit-side overrides), but the leak is masked because the parent sets `editRepo` to `null` on close, which causes `editFormDefaults` to recompute from scratch on the next open. Out of scope here; happy to file a follow-up if you'd like the same open-time reset pattern applied for symmetry.


brandonrc:
## Multi-agent review summary (3/3 reviewers)

**Reviewer 1 (Code Reviewer)**: APPROVE - correctly diagnoses the Radix programmatic-close gap and fixes it with an open-time useEffect plus a regression test that exercises the exact controlled-open path the bug hides in.

**Reviewer 2 (Frontend Developer)**: APPROVE - useEffect on [createOpen] correctly handles the programmatic-close path (verified parent calls setCreateOpen(false) in onSuccess, bypassing Radix onOpenChange); resetCreateForm only invokes stable useState setters so eslint-disable on exhaustive-deps is justified.

**Reviewer 3 (Evidence Collector)**: APPROVE - test correctly exercises the controlled-open path the author claims (rerenders createOpen true→false→true via props, bypassing onCreateOpenChange which is a vi.fn()); useEffect-on-open is the right minimal fix; verified pivotal at src/app/(app)/repositories/_components/repo-dialogs.test.tsx:333-356.

Verdict: ready to merge when CI green.

PR #422, opened 2026-05-20
fix(repositories): sort Format options alphabetically in Create dialog

dragonpaw:
## Summary

The Create Repository dialog's Format dropdown rendered options in the source-array order (grouped by ecosystem: Core, Container, Linux, …). In a flat single-list Select that grouping is invisible and reads as random, making it hard to find a format.

Fix: sort a local copy of `FORMAT_OPTIONS` by label at render time in `repo-dialogs.tsx`. The source array stays in ecosystem order so the grouped filter in `repositories-content.tsx` (which uses `FORMAT_GROUPS`) keeps its intentional headers, and `TYPE_OPTIONS` (which has an explicit order test) is left untouched.

## Test Checklist
- [x] Manually tested locally

Added a small regression test in `repo-dialogs.test.tsx` asserting the rendered Format option labels are sorted (case-insensitive).

## UI Changes
- [x] Dark mode verified
- [ ] N/A - no UI changes


brandonrc:
## Multi-agent review summary (3/3 reviewers)

**Reviewer 1 (Code Reviewer)**: APPROVE - sorts a local copy at module scope (cheap, runs once), leaves FORMAT_OPTIONS source order intact so the grouped filter elsewhere still works, and ships a case-insensitive sort assertion test.

**Reviewer 2 (Frontend Developer)**: APPROVE - uses [...FORMAT_OPTIONS].sort(...) so no in-place mutation of the imported constant, computed at module scope (no per-render cost), and grouped filter elsewhere is intentionally left untouched.

**Reviewer 3 (Evidence Collector)**: APPROVE - sorted-copy approach correctly preserves grouped filter in repositories-content.tsx, and the test compares rendered labels against [...labels].sort() which is genuinely pivotal because source order (Maven, Gradle, NPM, PyPI...) is non-alphabetical so a missed sort would fail it.

Verdict: ready to merge when CI green.

PR #447, opened 2026-06-01
feat(repo-browser): add 'Invalidate cache' action to artifact details dialog (closes #446)

### LINKED ISSUE #446: feat(repo-browser): add 'Invalidate cache' action to artifact details dialog (Remote repos)
## Problem Statement

When viewing the artifact-details dialog on a Remote (proxy) repository, there is no way for an operator to evict a single cached artifact entry. To force a re-fetch from upstream they currently have to either:

- delete the underlying blob in GCS / S3 out-of-band (requires storage-level credentials), or
- wait for the cache TTL to expire (configurable per-repo, default 24h).

Neither path is acceptable when the cache is known-bad — for example after a security scan flags a cached artifact, or when an upstream mirror serves corrupt content under a known path.

## Proposed Solution

Add an **"Invalidate cache"** action to the artifact-details dialog (the dialog opened from the `Info` tooltip / row click on the Artifacts tab in `repo-detail-content.tsx`). The action:

- Is only shown on artifact rows belonging to a `Remote` repository (cache invalidation is a no-op for Local / Virtual repos).
- Opens a small confirmation step (one-click destructive-ish action — same UX weight as the existing `Delete` action), then calls the backend.
- On success, toasts "Cache entry invalidated" and invalidates the React-Query cache for the artifacts list so a subsequent fetch goes back to upstream.
- On failure, surfaces the backend's user-message via the existing `mutationErrorToast` helper.

Backend dependency: companion endpoint `POST /api/v1/repositories/{key}/cache/invalidate?path=...` is being added in [artifact-keeper](https://github.com/artifact-keeper/artifact-keeper). The web change uses `apiFetch` (from `src/lib/api/fetch.ts`) until the SDK is regenerated, then can be migrated to the typed SDK call in a follow-up.

## Alternatives Considered

- **Row-level icon button** (next to Scan / Delete in the artifact table). Faster to invoke but less discoverable and easier to mis-click on a long list. The dialog placement is more deliberate, which matches the destructive intent of evicting a cache entry.
- **Repo-level "purge entire cache" button** in repo settings. Useful but a broader / scarier operation, out of scope for this issue.

## Use Case

As a repository admin, when I learn that a specific cached artifact in a remote repository is bad (security finding, upstream corruption, wrong tag, etc.), I want to click into the artifact's details and evict its cache entry so the next download re-fetches from upstream — without needing GCS / S3 credentials.

## Component

Repository Browser

## Additional Context

- Companion backend issue: [artifact-keeper#1539](https://github.com/artifact-keeper/artifact-keeper/issues/1539) (wires the HTTP endpoint this UI calls).
- The action belongs in `src/app/(app)/repositories/_components/repo-detail-content.tsx` inside the artifact-details `<Dialog>` block (around the existing tabs).
- API wrapper goes in `src/lib/api/artifacts.ts` alongside the existing `delete` / `getDownloadUrl` methods.

## Pre-submission Checklist

- [x] Reviewed the docs / UI to confirm this action does not already exist.

### LINKED ISSUE #1539: feat(api): wire HTTP endpoint for ProxyService::invalidate_cache (POST /repositories/:key/cache/invalidate?path=...)
## Problem Statement

`ProxyService::invalidate_cache` exists in the backend (`backend/src/services/proxy_service.rs`, around L1027) and is also covered by unit tests via the sister method `invalidate_cache_by_key` (path-traversal rejection, deletes both content + metadata blobs). However, **no HTTP handler wires it up**. Operators who need to evict a stale cached artifact from a remote (proxy) repository currently have to delete the underlying blob in object storage out-of-band (GCS / S3), which is awkward, easy to get wrong, and impossible to do without storage-level credentials.

## Proposed Solution

Wire up a new endpoint:

```
POST /api/v1/repositories/{key}/cache/invalidate?path=<artifact-path>
```

Behaviour:

- `key` (path) — repository key.
- `path` (query) — the artifact path to evict, identical to the path used by `GET /api/v1/repositories/{key}/artifacts/{path}`.
- Auth: requires authenticated user with `write` scope and `require_repo_access` on the target repo (mirrors `set_cache_ttl`).
- Repo type guard: only valid for `repo_type = Remote`. Other repo types return `400 Validation` (cache invalidation is meaningless on Local / Virtual repos — no proxy cache exists to evict).
- Storage guard: when `state.proxy_service` is `None` (e.g. storage backend not configured) returns `503 ServiceUnavailable` instead of crashing.
- On success, returns `200` with `{"repository_key": "...", "path": "...", "invalidated": true}`. The endpoint is idempotent — a second call for an already-evicted path still returns 200, matching the underlying `ProxyService::invalidate_cache` contract (it ignores delete-of-missing).
- `#[utoipa::path]` annotation so it appears in the generated OpenAPI spec / SDK.

The handler is a thin wrapper:

```rust
let proxy = state.proxy_service.as_ref().ok_or_else(|| {
    AppError::ServiceUnavailable("proxy service not configured".to_string())
})?;
proxy.invalidate_cache(&repo, &q.path).await?;
```

Path validation (rejection of `..` traversal, etc.) is already handled inside `cache_storage_key` and is covered by `test_invalidate_cache_by_key_rejects_invalid_path`. The new handler inherits that behaviour for free.

## Alternatives Considered

- **JSON body instead of query string** (`POST /:key/cache/invalidate` with `{"path": "..."}`). More consistent with `set_cache_ttl`, but the query form is simpler from the CLI / `curl` and is what the operator-facing UX is most likely to look like (paste a path, hit a button). Easy to accept either later if reviewers prefer the body form.
- **`DELETE /:key/cache?path=...`**. Semantically a delete, but the existing route shape under `/:key/cache-ttl` already uses `cache-*` as the noun; nesting `cache/invalidate` keeps cache-related operations under one prefix without colliding with `/:key/artifacts/*path` (which is the actual artifact resource).
- **Bulk invalidate by prefix**. Useful but a separate, larger discussion — out of scope for this issue.

## Use Case

As a repository admin operating a remote (proxy) repo, when upstream serves corrupt or wrong content under a known path (e.g. a tag was force-pushed, an upstream mirror went bad, a malware-scanner just flagged a cached artifact), I want to evict that single cache entry over the API so the next request re-fetches from upstream — without needing GCS / S3 credentials or a backend redeploy.

## Component

Backend / API

## Additional Context

- Existing method to wire: `ProxyService::invalidate_cache` (`backend/src/services/proxy_service.rs` L1027).
- Sibling endpoint to mirror style: `set_cache_ttl` / `get_cache_ttl` at `/{key}/cache-ttl`.
- Companion web-UI work tracked separately in [artifact-keeper-web](https://github.com/artifact-keeper/artifact-keeper-web) — an "Invalidate cache" action on the artifact-details dialog. The web change calls this endpoint.

## Pre-submission Checklist

- [x] Searched existing issues / PRs.
- [x] Reviewed the docs to confirm there is no existing endpoint.

knowinglyAnonymous:
## Summary

Adds an **"Invalidate cache"** action to the artifact-details dialog on Remote (proxy) repositories so operators can evict a single cached artifact entry from the UI instead of having to delete blobs out-of-band in object storage.

Closes #446.

The button calls the new backend endpoint `POST /api/v1/repositories/{key}/cache/invalidate?path=...` (companion PR [artifact-keeper#1540](https://github.com/artifact-keeper/artifact-keeper/pull/1540), tracked by [artifact-keeper#1539](https://github.com/artifact-keeper/artifact-keeper/issues/1539)). Routed through `apiFetch` since the generated SDK has not yet been regenerated against the new endpoint — once it has, the wrapper can collapse to the typed SDK call in a follow-up.

### What changes

- `src/lib/api/artifacts.ts` — new `artifactsApi.invalidateCache(repoKey, path)` wrapper. URL-encodes both segments so `+`, `/`, and spaces in artifact paths survive the round-trip back to the backend.
- `src/app/(app)/repositories/_components/repo-detail-content.tsx` — new React-Query mutation `invalidateCacheMutation`. On success it invalidates `["artifacts", repoKey]` and `["repository", repoKey]` so the next list/detail fetch goes back to upstream, and toasts a clear "next download will re-fetch from upstream" message. On failure surfaces the backend's user-message via `mutationErrorToast`.
- New `<Button variant="outline">` with the `RotateCcw` icon in the dialog footer, **gated on `repository.repo_type === "remote"`**. Cache invalidation is meaningless for Local / Virtual / Staging repos so the action is hidden entirely there. The backend also rejects with 400 for non-remote repos, so this is defence in depth.
- Pending state shows "Invalidating..." text + disables the button to prevent double-fires.

### What does *not* change

- The Delete and Download buttons are untouched. Invalidate-cache is intentionally a separate action — it removes the cached copy but does not remove the artifact row from the database, so a follow-up download repopulates the cache from upstream. That's the whole point of the operator workflow this enables.
- No SDK regeneration is required to land this PR; once the backend endpoint is in the OpenAPI spec the wrapper can move to the typed SDK call.

## Test Checklist

- [x] Unit tests added/updated — three new vitest cases for `invalidateCache` (`src/lib/api/__tests__/artifacts.test.ts`):
  - `issues POST to /api/v1/repositories/:key/cache/invalidate?path=...` — pins URL, method, credentials.
  - `URL-encodes both the repository key and the path` — protects against operator mistakes with weird repo keys / artifact paths containing `+`, `/`, spaces.
  - `throws on non-ok response (e.g. 400 for non-remote repo, 503 for missing storage)` — pins error surfacing.
- [ ] E2E Playwright tests added/updated — Deferred. The button only renders for Remote repos, so a meaningful E2E needs a working remote-repo proxy fixture in the test harness. Happy to add one as a follow-up if reviewers want it; the unit test plus the explicit `repo_type === "remote"` guard cover the visibility contract.
- [x] Manually verified via `npm run lint` (0 errors), `npx tsc --noEmit` (no new errors; 2 pre-existing on `main` unchanged), `npm run build` (clean), `npx vitest run src/lib/api` (587/587 pass).

## UI Changes

- [x] Responsive layout verified — the new button sits in the same `<DialogFooter>` row as Delete / Download, inheriting the existing flex behaviour. No new layout primitives.
- [x] Dark mode verified — uses `variant="outline"` with no custom colours, so it inherits theme tokens like the existing footer buttons.
- [x] Accessibility checked — keyboard-focusable like the surrounding buttons, includes a `title` tooltip explaining the action ("Evict this artifact from the proxy cache; next download re-fetches from upstream"), and the icon-plus-text pattern matches the existing Delete and Download buttons (no icon-only buttons added).
- [ ] N/A - no UI changes

## Companion work

- Backend endpoint that this UI calls: [artifact-keeper#1540](https://github.com/artifact-keeper/artifact-keeper/pull/1540) (closes [artifact-keeper#1539](https://github.com/artifact-keeper/artifact-keeper/issues/1539)). The frontend works locally as soon as that endpoint ships; the React-Query wiring here is independent of SDK regeneration.

### DIFF
--- a/src/app/(app)/repositories/_components/repo-detail-content.tsx
+++ b/src/app/(app)/repositories/_components/repo-detail-content.tsx
@@ -42,6 +42,7 @@ import { MavenComponentList } from "./maven-component-list";
 import { DockerTagList } from "./docker-tag-list";
 import { QuarantineBadge } from "@/components/common/quarantine-badge";
 import { QuarantineBanner } from "@/components/common/quarantine-banner";
+import { ConfirmDialog } from "@/components/common/confirm-dialog";
 import { RepoSettingsTab } from "./repo-settings-tab";
 import { formatBytes, REPO_TYPE_COLORS } from "@/lib/utils";
 import { useAuth } from "@/providers/auth-provider";
@@ -113,6 +114,11 @@ export function RepoDetailContent({ repoKey, standalone = false }: RepoDetailCon
   // artifact detail dialog
   const [detailOpen, setDetailOpen] = useState(false);
   const [selectedArtifact, setSelectedArtifact] = useState<Artifact | null>(null);
+  // Confirm gate for the destructive "Invalidate cache" action (#446 review).
+  const [invalidateConfirmOpen, setInvalidateConfirmOpen] = useState(false);
+  // Polite live-region text announcing the cache-invalidation outcome to
+  // assistive tech (#446 review): the toast alone is not reliably announced.
+  const [cacheActionStatus, setCacheActionStatus] = useState("");
 
   // security form local state
   const [secForm, setSecForm] = useState<UpsertScanConfigRequest | null>(null);
@@ -218,15 +224,33 @@ export function RepoDetailContent({ repoKey, standalone = false }: RepoDetailCon
   // for repos without a cache.
   const invalidateCacheMutation = useMutation({
     mutationFn: (path: string) => artifactsApi.invalidateCache(repoKey, path),
-    onSuccess: () => {
+    onSuccess: async (_result, path) => {
       // Drop the artifacts list and repo summary from the cache so the next
       // fetch goes back to upstream (the underlying download endpoint will
       // re-populate the proxy cache on the next access).
       queryClient.invalidateQueries({ queryKey: ["artifacts", repoKey] });
       queryClient.invalidateQueries({ queryKey: ["repository", repoKey] });
+      // The dialog holds a local copy of the artifact; after eviction its
+      // cache metadata (cache_cached_at / cache_expires_at, surfaced by #449)
+      // is stale. Refresh the open row from the server so those values
+      // reflect the eviction instead of the pre-eviction snapshot (#446
+      // review). Best-effort: the list/summary are already invalidated, so a
+      // failed refresh just leaves the last-known values in place.
+      try {
+        const fresh = await artifactsApi.get(repoKey, path);
+        setSelectedArtifact((prev) => (prev && prev.path === path ? fresh : prev));
+      } catch {
+        // non-fatal
+      }
+      setCacheActionStatus(
+        "Cache entry invalidated; the next download will re-fetch from upstream.",
+      );
       toast.success("Cache entry invalidated; next download will re-fetch from upstream.");
     },
-    onError: mutationErrorToast("Failed to invalidate cache"),
+    onError: (err: unknown) => {
+      setCacheActionStatus("Cache invalidation failed. Please try again.");
+      mutationErrorToast("Failed to invalidate cache")(err);
+    },
   });
 
   const scanRepoMutation = useMutation({
@@ -290,6 +314,7 @@ export function RepoDetailContent({ repoKey, standalone = false }: RepoDetailCon
 
   const showDetail = useCallback((artifact: Artifact) => {
     setSelectedArtifact(artifact);
+    setCacheActionStatus("");
     setDetailOpen(true);
   }, []);
 
@@ -979,11 +1004,7 @@ export function RepoDetailContent({ repoKey, standalone = false }: RepoDetailCon
                 {repository.repo_type === "remote" && (
                   <Button
                     variant="outline"
-                    onClick={() => {
-                      if (selectedArtifact) {
-                        invalidateCacheMutation.mutate(selectedArtifact.path);
-                      }
-                    }}
+                    onClick={() => setInvalidateConfirmOpen(true)}
                     disabled={invalidateCacheMutation.isPending}
                     title="Evict this artifact from the proxy cache; next download re-fetches from upstream"
                   >
@@ -1000,8 +1021,37 @@ export function RepoDetailContent({ repoKey, standalone = false }: RepoDetailCon
               </>
             )}
           </DialogFooter>
+          {/*
+            Polite live region for async cache-invalidation feedback (#446
+            review). Toasts are not reliably announced by assistive tech, and
+            the button's "Invalidating..." label swap is otherwise silent.
+            `sr-only` keeps it visually invisible while announcing to AT.
+          */}
+          <div role="status" aria-live="polite" className="sr-only">
+            {invalidateCacheMutation.isPending
+              ? "Invalidating cache entry…"
+              : cacheActionStatus}
+          </div>
         </DialogContent>
       </Dialog>
+
+      {/* Confirm gate for the destructive "Invalidate cache" action (#446). */}
+      <ConfirmDialog
+        open={invalidateConfirmOpen}
+        onOpenChange={setInvalidateConfirmOpen}
+        title="Invalidate cached artifact?"
+        description="This evicts the cached copy from the proxy. The next download re-fetches it from the upstream remote. This cannot be undone."
+        confirmText="Invalidate cache"
+        danger
+        loading={invalidateCacheMutation.isPending}
+        onConfirm={() => {
+          if (selectedArtifact) {
+            invalidateCacheMutation.mutate(selectedArtifact.path, {
+              onSettled: () => setInvalidateConfirmOpen(false),
+            });
+          }
+        }}
+      />
     </div>
   );
 }
--- a/src/app/(app)/repositories/_components/repo-detail-content.tsx
+++ b/src/app/(app)/repositories/_components/repo-detail-content.tsx
@@ -17,6 +17,7 @@ import {
   Layers,
   Package as PackageIcon,
   Settings,
+  RotateCcw,
 } from "lucide-react";
 
 import { repositoriesApi } from "@/lib/api/repositories";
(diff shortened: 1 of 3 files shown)


brandonrc (commented):
Accessibility review (WCAG 2.2 AA) — repo-detail-content.tsx Invalidate cache button.

1. Destructive action has no confirmation (the PR title says "+ confirm", but the diff fires `invalidateCacheMutation.mutate()` directly from onClick). Cache invalidation is a destructive, hard-to-undo proxy operation. Wrap it in the same AlertDialog confirm pattern the Delete action uses, so keyboard and screen-reader users get an explicit confirm step rather than a single-click commit. (3.3.4 Error Prevention — for destructive actions; consistency with the adjacent Delete button.)

2. Async result is toast-only — no live region (4.1.3 Status Messages). On success/failure the only feedback is `toast.success`/`mutationErrorToast`. Sonner toasts are not consistently announced to AT, and the button's "Invalidating..." -> done transition is a silent text swap. Add an in-dialog `role="status"` (aria-live="polite") region that receives the outcome text, mirroring the live-region pattern in #459 (`upstream-auth-status`).

3. Minor — pending state: while `disabled` + label changes to "Invalidating...", the label change alone is not announced. The live region above covers this; alternatively add `aria-live` to the label span.

Accessible name on the button itself is fine (visible text "Invalidate cache" accompanies the RotateCcw icon). The `title` tooltip is a nice-to-have but should not be the only place the action's effect is described.

Not blocking the logic, but item 1 (missing confirm) is a behavior gap vs. the PR description and should be resolved before merge.

brandonrc (commented):
Frontend correctness review. Overall this is clean and well-scoped: the mutation wiring, cache invalidation, and the `repo_type === "remote"` gate are all correct, and the test coverage (URL/encoding, non-ok throw, idempotency note) is genuinely good.

No blockers. A couple of suggestions:

🟡 **Stale `selectedArtifact` after invalidation (repo-detail-content.tsx, onSuccess)**
On success you `invalidateQueries(["artifacts", repoKey])` and `["repository", repoKey]`, which is right for the list, but the details dialog keeps rendering the old `selectedArtifact` object held in local state. After eviction the cache_cached_at / cache_expires_at rows (added in #451) would still show the pre-eviction values until the user reopens the dialog. Consider either closing the dialog on success or refreshing the selected artifact from the refetched list. Not a correctness bug today, but it becomes a visible inconsistency once #451 lands.

💭 **Button is enabled even when `selectedArtifact` is null**
The onClick guards with `if (selectedArtifact)`, so it is safe, but the button has no disabled state tied to selection. Since it only renders inside the details dialog where an artifact is always selected, this is cosmetic.

Note (not a defect): this touches `repo-detail-content.tsx`, which #451 / #458 / #460 / #462 / #464 also modify. The button block and the cache-row block are in different regions of the JSX, so conflicts should be mechanical, but flag merge order to whoever sequences the wave.

brandonrc (commented):

The web side routes `POST /api/v1/repositories/{key}/cache/invalidate?path=...` via `apiFetch` with `credentials: 'include'`. Authorization and the "remote repos only" rule are enforced server-side (the PR notes the backend rejects non-remote repos with 400 and the comment references `ProxyService::invalidate_cache`); the client `repo_type === "remote"` gate is correctly only a UX affordance, not the security boundary. Good that it does not pretend to be authz.

Injection: both `repoKey` and `artifactPath` are `encodeURIComponent`-encoded, and `path` is carried in the query string. The test pins that `/` and `+` survive as `%2F`/`%2B`, so the segment cannot break out of the query into the path or inject extra query params. No path-traversal vector reaches a different repo's cache from the client construction.

CSRF: this is a state-changing POST relying on `credentials: 'include'` (cookie auth). Whether this is exploitable depends entirely on the backend's CSRF posture for cookie-authenticated mutations. This PR does not introduce a new gap relative to the other mutations in this codebase (scan trigger, delete, etc. follow the same pattern), but flagging for the team: confirm the API enforces a non-cookie auth requirement (bearer token / custom header / SameSite=strict + CSRF token) on these admin POSTs. If the existing mutations are safe, this one inherits that safety. Non-blocking for this PR specifically since it matches the established pattern.

Permission gating: the invalidate action is rendered in the repo detail view; ensure the backend requires repo-admin/write permission (not merely authenticated read) so a read-only user cannot evict cache and force upstream re-fetches (a mild DoS / cache-busting lever). Verify server-side; not visible in this diff.

No blocking issues on the web change. One thing to confirm with the backend team: write-permission gating + CSRF posture on the new endpoint.

brandonrc:
Thank you for this contribution. Your work shipped in **v1.2.0**: the Invalidate cache action on the artifact details dialog feature was incorporated into the release integration branch #466, which was squash-merged to main and tagged as [v1.2.0](https://github.com/artifact-keeper/artifact-keeper-web/releases/tag/v1.2.0), resolving #446. The feature is live on main now.

PR #450, opened 2026-06-01
feat(settings): show + edit proxy cache TTL on Remote repo Settings tab (closes #448)

### LINKED ISSUE #448: feat(settings): show + edit proxy cache TTL on Remote repo Settings tab
## Problem Statement

The proxy cache TTL for a Remote (proxy) repository is configurable over the API (`PUT /api/v1/repositories/{key}/cache-ttl`) and readable (`GET /api/v1/repositories/{key}/cache-ttl`), but the web Settings tab does not surface either. Operators currently have to use `curl` (or read the docs that PR #71 fixed) to inspect or change the TTL — there is no in-UI affordance.

## Proposed Solution

Add a **"Proxy Cache"** section to the existing Settings tab (`src/app/(app)/repositories/_components/repo-settings-tab.tsx`), shown only when `repository.repo_type === "remote"`:

- A read-only "Currently configured TTL" line, populated from `GET /:key/cache-ttl`. The endpoint already falls back to `DEFAULT_CACHE_TTL_SECS` (24h) when nothing is stored, so the value displayed is what the proxy will actually apply — matching the contract the doc PR #71 spelled out.
- An editable input for the new TTL value (in seconds, with a small helper showing the equivalent in minutes/hours/days). Backend validates the range as `1..=2_592_000` (1s to 30d); the UI mirrors that and surfaces a clear inline error for out-of-range values.
- The new field plugs into the existing form's `hasChanges` / "Save Changes" / "Discard" workflow already used for name / description / quota, so operators get one consistent save bar at the bottom rather than yet another submit button.
- For non-Remote repos the section is hidden entirely. Backend rejects `PUT /:key/cache-ttl` on non-Remote with 400 anyway (see `is_cache_ttl_configurable`); hiding the section avoids letting users build invalid changes that fail at save time.

API wrappers (`getCacheTtl`, `setCacheTtl`) go on `repositoriesApi` in `src/lib/api/repositories.ts`. The generated SDK already exports `getCacheTtl` and `setCacheTtl`, so this is a typed-SDK call — no hand-written `apiFetch` needed.

## Alternatives Considered

- **Standalone "Cache" tab on the repo detail page.** Discoverable but heavyweight for a single integer; the Settings tab already collects all repo-level configuration so a section is a better fit.
- **Inline editor on the repo header / breadcrumb area.** Too noisy; cache TTL is an admin knob, not a primary navigation surface.
- **Read-only display only, edits via API only.** Defeats the point of the issue — we want one-click edits without leaving the UI.

## Use Case

As a repository admin operating a remote (proxy) repo, when I notice upstream is publishing fresh artifacts faster than my current 24h default cache lets me serve, I want to inspect the configured TTL and shorten it (e.g. to 1h) directly from the repo's Settings tab — without dropping to `curl` or reading docs.

## Component

Settings

## Additional Context

- Backend endpoints already exist and the SDK already exposes them — this is a pure UI change.
- Companion related work: the per-artifact "Invalidate cache" button shipped via [#446](https://github.com/artifact-keeper/artifact-keeper-web/issues/446) / [#447](https://github.com/artifact-keeper/artifact-keeper-web/pull/447) covers eviction of single entries; this issue is the repo-wide TTL knob.

## Pre-submission Checklist

- [x] Reviewed the docs / UI to confirm the section does not already exist.

knowinglyAnonymous:
## Summary

Adds a **"Proxy Cache"** section to the repo Settings tab that displays and edits the proxy cache TTL on Remote (proxy) repositories. Hidden for Local / Virtual / Staging — those repo types have no proxy cache.

Closes #448.

The backend endpoints (`GET` / `PUT /api/v1/repositories/{key}/cache-ttl`) already exist and the generated SDK already exports `getCacheTtl` / `setCacheTtl`, so this is a pure UI change — no backend or SDK regeneration needed.

### What changes

- `src/lib/api/repositories.ts` — adds typed `repositoriesApi.getCacheTtl(repoKey)` and `repositoriesApi.setCacheTtl(repoKey, seconds)` wrappers around the SDK calls.
- `src/app/(app)/repositories/_components/repo-settings-tab.tsx` — new "Proxy Cache" section shown only when `repository.repo_type === "remote"`. Plugs into the existing `hasChanges` / "Save Changes" / "Discard" workflow so operators get one consistent save bar at the bottom of the tab rather than yet another submit button.
- The TTL update goes through a separate `setCacheTtlMutation` because the backend exposes it as a distinct endpoint from `update_repository`. `handleSave` now dispatches both mutations in parallel via `Promise.allSettled` — a 4xx on one side does not roll back the other, and each mutation already wires its own `onError` toast so the operator sees exactly which side failed.
- Inline validation mirrors the backend's `validate_cache_ttl` range (`1..=2_592_000`, i.e. 1s to 30d): the input is `aria-invalid` when out of range, the Save button is disabled until the value is valid, and the helper line shows the human-readable equivalent (`≈ 24 hours`, `≈ 1 day 6 hours`) so operators don't have to compute "is 86400 sensible?" in their head.
- Default values come from `GET /:key/cache-ttl`, which is documented to return the effective default (`86400` = 24h) when nothing is stored — matching the contract the docs PR #71 spelled out.

### What does *not* change

- Non-Remote repos: the section is hidden entirely. `is_cache_ttl_configurable` already rejects `PUT /:key/cache-ttl` for those types, so hiding the section avoids letting users compose changes that fail at save time.
- Per-artifact cache invalidation continues to live on the artifact-details dialog ([#446](https://github.com/artifact-keeper/artifact-keeper-web/issues/446) / [#447](https://github.com/artifact-keeper/artifact-keeper-web/pull/447)). This PR is the repo-wide knob; the per-artifact button is the targeted scalpel.

## Test Checklist

- [x] Unit tests added/updated — five new vitest cases in `src/lib/api/__tests__/repositories.test.ts` covering:
  - `getCacheTtl` happy path + error surfacing.
  - `setCacheTtl` happy path with body-shape pin (matches `SetCacheTtlRequest`, not the legacy `value` form the docs PR #71 corrected).
  - `setCacheTtl` error surfacing including the backend's "remote (proxy) repositories" rejection message for non-Remote repos.
- [ ] E2E Playwright tests added/updated — Deferred. Adding meaningful E2E coverage for the section requires a Remote-repo fixture in the harness; the unit tests plus the explicit `repo_type === "remote"` gate cover the visibility contract.
- [x] Manually verified via `npm run lint` (0 errors, 36 pre-existing warnings unchanged), `npx tsc --noEmit` (no new errors; 2 pre-existing on `main` unchanged), `npm run build` (clean), `npx vitest run src/lib/api/__tests__/repositories.test.ts` (18/18 pass; was 13 before).

## UI Changes

- [x] Responsive layout verified — the new `<section>` follows the existing `max-w-2xl space-y-8` layout used by General / Storage / Cleanup Policies.
- [x] Dark mode verified — uses theme tokens (`text-muted-foreground`, `text-destructive`) with no custom colours, inheriting the rest of the tab.
- [x] Accessibility checked — Label is associated via `htmlFor`, `aria-invalid` is set on the input when the value is out of range, the inline error and helper text are visible to screen readers in document order, and the Save button is disabled while the value is invalid.
- [ ] N/A - no UI changes

### DIFF
--- a/src/app/(app)/repositories/_components/repo-settings-tab.test.tsx
+++ b/src/app/(app)/repositories/_components/repo-settings-tab.test.tsx
@@ -905,6 +905,75 @@ describe("RepoSettingsTab - Proxy Cache section (#448)", () => {
     expect(save).toHaveProperty("disabled", true);
   });
 
+  it("programmatically associates the validation error with the input (#448 a11y)", async () => {
+    const user = userEvent.setup();
+    render(<RepoSettingsTab repository={remoteRepo} />, {
+      wrapper: createWrapper(),
+    });
+    await waitFor(() => {
+      expect(screen.getByLabelText("Cache TTL (seconds)")).toHaveProperty(
+        "value",
+        "86400"
+      );
+    });
+
+    const input = screen.getByLabelText("Cache TTL (seconds)");
+    await user.clear(input);
+    await user.type(input, "9999999");
+
+    // The error must be exposed as a live alert and linked from the input
+    // via aria-describedby so screen-reader users hear the explanation, not
+    // just "invalid". Mirrors the reference pattern in #459.
+    const error = screen.getByRole("alert");
+    expect(error.id).toBe("settings-cache-ttl-error");
+    expect(input.getAttribute("aria-describedby")).toBe(
+      "settings-cache-ttl-error"
+    );
+
+    // When the value becomes valid again, the association is removed.
+    await user.clear(input);
+    await user.type(input, "3600");
+    expect(screen.queryByRole("alert")).toBeNull();
+    expect(input.getAttribute("aria-describedby")).toBeNull();
+  });
+
+  it("disables Discard while a cache-TTL save is in flight (#448 review)", async () => {
+    // A never-resolving setCacheTtl keeps the mutation pending so we can
+    // assert the Discard button is guarded the same way Save is, preventing
+    // a Discard from racing an in-flight save.
+    let resolveSet: (() => void) | undefined;
+    mockSetCacheTtl.mockReturnValue(
+      new Promise<void>((resolve) => {
+        resolveSet = resolve;
+      })
+    );
+
+    const user = userEvent.setup();
+    render(<RepoSettingsTab repository={remoteRepo} />, {
+      wrapper: createWrapper(),
+    });
+    await waitFor(() => {
+      expect(screen.getByLabelText("Cache TTL (seconds)")).toHaveProperty(
+        "value",
+        "86400"
+      );
+    });
+
+    const input = screen.getByLabelText("Cache TTL (seconds)");
+    await user.clear(input);
+    await user.type(input, "3600");
+    await user.click(screen.getByRole("button", { name: /save changes/i }));
+
+    await waitFor(() => {
+      expect(
+        screen.getByRole("button", { name: /discard/i })
+      ).toHaveProperty("disabled", true);
+    });
+
+    // Let the in-flight mutation settle so the test doesn't leak a pending promise.
+    resolveSet?.();
+  });
+
   it("zero is rejected as out-of-range (backend min is 1)", async () => {
     const user = userEvent.setup();
     render(<RepoSettingsTab repository={remoteRepo} />, {
--- a/src/app/(app)/repositories/_components/repo-settings-tab.test.tsx
+++ b/src/app/(app)/repositories/_components/repo-settings-tab.test.tsx
@@ -37,9 +37,13 @@ vi.mock("sonner", () => ({
 
 // Mock repositories API
 const mockUpdate = vi.fn();
+const mockGetCacheTtl = vi.fn();
+const mockSetCacheTtl = vi.fn();
 vi.mock("@/lib/api/repositories", () => ({
   repositoriesApi: {
     update: (...args: unknown[]) => mockUpdate(...args),
+    getCacheTtl: (...args: unknown[]) => mockGetCacheTtl(...args),
+    setCacheTtl: (...args: unknown[]) => mockSetCacheTtl(...args),
   },
 }));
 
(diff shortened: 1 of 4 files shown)


brandonrc (commented):
Accessibility review (WCAG 2.2 AA) — repo-settings-tab.tsx Proxy Cache TTL field.

Solid baseline: the field has a real `<Label htmlFor="settings-cache-ttl">`, native `type="number"` with min/max/step, the section is wrapped in `<section aria-labelledby="settings-cache-heading">`, and `aria-invalid` is toggled correctly. One real gap:

1. Validation error is not programmatically associated with the input (3.3.1 Error Identification / 4.1.3 Status Messages / 1.3.1). The error text:

       <p className="text-sm text-destructive">Must be a whole number between 1 and 2,592,000.</p>

   is set `aria-invalid` on the input but never linked to it, and carries no `role="alert"`. A screen-reader user who focuses the field hears "invalid" with no explanation, and the message is not announced when it appears. This diverges from the repo reference pattern in #459 (input `aria-describedby` -> error element `id`, error element `role="alert"`). Fix:
   - Give the error `<p>` a stable `id` (e.g. `settings-cache-ttl-error`) and `role="alert"`.
   - Add `aria-describedby={cacheTtlOverride !== undefined && !cacheTtlIsValid ? "settings-cache-ttl-error" : undefined}` to the Input.

2. Minor — the `≈ formatTtlHint(...)` helper line is decorative/derived and fine visually, but consider `aria-describedby` linking the persistent "Range: ..." hint to the input as well so its constraints are announced on focus (the native min/max already help, so this is optional).

Save outcome is toast-only here too (`toast.success("Cache TTL saved")`); consistent with the rest of the settings tab, so not raising it as new for this PR, but worth a follow-up for 4.1.3 across the settings tab.

Item 1 should be fixed to match #459 before merge.

brandonrc (commented):
Frontend correctness review. Solid PR. The independent-mutation design (general fields vs cache TTL as two separate endpoints, dispatched in parallel and awaited via `Promise.allSettled` so a 4xx on one side doesn't surface as an unhandled rejection) is the right call and well-documented. Validation mirrors the backend range, the `enabled: isRemote` gate avoids wasted GETs, and the controlled-string-input approach for mid-edit state is correct. Test coverage is thorough.

No blockers. Suggestions:

🟡 **`handleSave` resolves before the user sees the result, and Discard can race an in-flight save (repo-settings-tab.tsx)**
`handleSave` awaits `Promise.allSettled`, but nothing prevents Discard being clicked while the mutations are pending; `handleDiscard` clears `cacheTtlOverride` and on a subsequent `setCacheTtl` onSuccess you also clear it. Low risk, but worth confirming the Discard button is disabled during `setCacheTtlMutation.isPending` the same way Save is. Right now Discard has no pending guard.

🟡 **`formatTtlHint` drops a trailing seconds component for mixed values**
For e.g. 90061s (1 day 1 hour 1 minute 1 second) the `secs` branch only pushes when `parts.length === 0`, so the seconds are silently omitted. That is an intentional compactness choice for the common case, but the hint then slightly misrepresents the exact value. Since the field accepts arbitrary seconds, consider noting "approximately" (the `≈` prefix already implies this, so this is borderline a nit).

💭 **`cacheTtlChanged` compares `parsedCacheTtl !== currentCacheTtlSeconds` where parsed can be `NaN`**
If the user clears then types a non-numeric value, `Number("")` is handled (null branch) but `Number("1e")` style inputs on a number input are generally blocked by the browser. `cacheTtlIsValid` gates the actual save, so this is safe; just calling out that `cacheTtlChanged` can be true with an invalid value, which is exactly why the Save disable condition also checks `!cacheTtlIsValid`. Correct as written.

Note (not a defect): touches `repo-settings-tab.tsx`, which #452 and the other wave-1 PRs also modify (#452 inserts an `AgeGateSettings` section into the same render). Conflicts are likely but mechanical. Flag merge order.

brandonrc (commented):

Input validation: solid. The component pins `CACHE_TTL_MIN_SECONDS = 1` and `CACHE_TTL_MAX_SECONDS = 2_592_000` to mirror the backend `validate_cache_ttl` range, and `cacheTtlIsValid` requires `Number.isInteger` plus the bounds. Save is disabled and `aria-invalid` is set for out-of-range / non-integer / zero / empty input, so a malformed value never reaches `setCacheTtl`. Backend remains the authoritative validator (the comment correctly frames the client check as a UX improvement over an opaque 400), which is the right layering.

Admin gating: the section is gated on `repo_type === "remote"`, and the `getCacheTtl` query is `enabled: isRemote` so no wasted/forbidden round-trip fires for Local/Virtual/Staging. The settings tab itself lives under the repository management surface; actual write authorization is enforced server-side on the TTL endpoint. No client-side privilege assumption is being relied on for security.

Save flow: TTL update and the general-fields update are dispatched as independent endpoints via `Promise.allSettled`, so a 4xx on one does not surface as an unhandled rejection and each has its own error toast. No empty PATCH is sent when only TTL changed (test pins `mockUpdate` not called) which keeps the audit log clean.

No injection surface (numeric value, typed endpoint). No sensitive-data exposure.

No blocking issues.

brandonrc:
Thank you for this contribution. Your work shipped in **v1.2.0**: the proxy cache TTL view/edit on the Remote repo Settings tab feature was incorporated into the release integration branch #466, which was squash-merged to main and tagged as [v1.2.0](https://github.com/artifact-keeper/artifact-keeper-web/releases/tag/v1.2.0), resolving #448. The feature is live on main now.

PR #451, opened 2026-06-01
feat(repo-browser): show artifact cache cached_at + expires_at in details dialog (closes #449)

### LINKED ISSUE #449: feat(repo-browser): show artifact cache cached_at + expires_at on details dialog (Remote repos)
## Problem Statement

The artifact-details dialog (opened from the Artifacts tab on a repo detail page) shows static fields like Name, Path, Size, Content Type, and Created. For a Remote (proxy) repository it does NOT show **when the cached copy of the artifact will expire**, which is exactly the field operators need before deciding whether to wait for the natural refresh or hit the "Invalidate cache" button shipped via [#446](https://github.com/artifact-keeper/artifact-keeper-web/issues/446) / [#447](https://github.com/artifact-keeper/artifact-keeper-web/pull/447).

## Proposed Solution

Once the backend exposes the per-artifact cache metadata (companion issue: artifact-keeper#... — backend issue link will be added in the PR description once filed), surface those fields in the existing artifact-details dialog `details` tab in `repo-detail-content.tsx`:

- New `DetailRow` for **"Cached"** (relative time + absolute timestamp tooltip), populated from `cache_cached_at`. Hidden when the value is absent.
- New `DetailRow` for **"Cache expires"** with relative-time formatting (e.g. "in 4 hours" or "expired 12 minutes ago, will re-fetch on next download"). Hidden when the value is absent.
- Both rows only render when `repository.repo_type === "remote"` AND the backend response actually included the fields. Local / Virtual / Staging repos do not have a proxy cache, so the rows never appear there.
- Sits next to the new "Invalidate cache" button so the operator can read the expiry state and act on it in the same panel.

The existing `Artifact` local type (`src/types/index.ts` or wherever it lives) gets two new optional fields mirroring the backend's optional response fields. The `adaptArtifact` helper in `src/lib/api/artifacts.ts` plumbs them through.

## Alternatives Considered

- **Inline badge on the artifact row in the table.** Cluttered for long lists; the dialog is the right place for this level of detail.
- **Polling the cache state every N seconds while the dialog is open.** Premature optimisation — operators reading this panel are deciding whether to invalidate, so a stale-by-a-minute display is fine. Can be revisited if real usage shows otherwise.
- **Show only `expires_at` and not `cached_at`.** Tempting, but having both makes "this entry is fresh" vs. "this entry is about to flip stale" much clearer, especially for short-TTL repos.

## Use Case

As a repo admin, when I open the details for a cached artifact in a remote (proxy) repository, I want to immediately see "Cache expires in 4 hours" so I can choose between waiting for the natural refresh and clicking "Invalidate cache" right now. Today the dialog gives me no signal at all about cache freshness.

## Component

Repository Browser

## Additional Context

- Hard dependency: the backend issue ([artifact-keeper#1541](https://github.com/artifact-keeper/artifact-keeper/issues/1541)) must merge first so the SDK regenerates with the new optional fields. This PR will block until then; a draft branch can land in parallel using hand-typed extensions to the `Artifact` type.
- Related shipped work in the same dialog: the per-artifact "Invalidate cache" button (#446 / #447). This issue completes the "see + act" loop alongside it.

## Pre-submission Checklist

- [x] Reviewed the docs / UI to confirm these rows do not already exist.

knowinglyAnonymous:
## Summary

Surfaces the proxy-cache `cached_at` and `expires_at` timestamps as relative-time rows on the artifact details dialog, so operators reading the panel can decide between waiting for the natural refresh and clicking "Invalidate cache" without leaving the UI.

Closes #449. Companion backend work tracked in artifact-keeper#1541 (PR artifact-keeper#1542).

### What changes

- **`src/lib/cache-time.ts`** (new) — two pure helpers, both `(iso, now?) => string`:
  - `formatRelativeTimestamp(iso)` → "in 4 hours" / "12 minutes ago" / "expired 3 days ago". Picks the largest unit that gives a magnitude `>= 1` so the output stays compact (no "120 minutes ago"). Uses `Intl.RelativeTimeFormat` so the output is locale-aware without pulling in `date-fns`.
  - `formatCacheExpiry(iso)` → biased for the "expires" framing; past timestamps render as "expired N units ago, will re-fetch on next download" so operators reading the row know what'll happen without checking docs.
  - Both helpers accept an explicit `now` parameter so the tests can pin output deterministically; production callers omit it.
- **`src/types/index.ts`** — `Artifact` gains `cache_cached_at?: string | null` and `cache_expires_at?: string | null`. Optional + nullable so the type tolerates the backend's `skip_serializing_if = "Option::is_none"` shape (key omitted entirely) and the defensive case of an explicit `null`.
- **`src/lib/api/artifacts.ts`** — `adaptArtifact` plumbs the new fields through. The generated SDK type doesn't carry them yet (the SDK regenerates from the OpenAPI spec only after the backend PR lands); until then they ride through via a narrowed runtime cast. Once the SDK regenerates, the cast can collapse to direct property access — comment in the file points this out.
- **`src/app/(app)/repositories/_components/repo-detail-content.tsx`** — adds two new `DetailRow`s in the dialog "Details" tab (between "Created" and "SHA-256"):
  - **Cached** — `formatRelativeTimestamp(cache_cached_at)` as the visible value, full ISO-8601 timestamp as the hover-tooltip.
  - **Cache expires** — `formatCacheExpiry(cache_expires_at)`, same tooltip pattern.
  - Both rows are gated on `repository.repo_type === "remote"` AND the value being present. Local / Virtual / Staging never show them; Remote repos without a cache-metadata blob (e.g. an artifact direct-uploaded but never proxy-fetched) also hide the rows rather than rendering empty cells.
- `DetailRow` gains an optional `title` prop so the visible value (e.g. "in 4 hours") can differ from the hover-tooltip (e.g. "6/2/2026, 11:00:00 AM"). Defaults to the value, so existing callers are unaffected.

### What does *not* change

- Listing / table view: the new fields are NEVER fetched into the table view. They only land via the per-artifact metadata response (matching the backend's cost-control choice in artifact-keeper#1542 — listings stay zero-storage).
- Polling: the dialog reads the value once when it opens; no auto-refresh while open. Operators who want a fresh value can close + reopen the dialog. Premature optimisation otherwise.

## Test Checklist

- [x] **Unit tests added/updated** —
  - `src/lib/__tests__/cache-time.test.ts` (new) — 9 cases covering: future direction, past direction, largest-unit selection (days, not 36 hours), unparseable-timestamp fallback, and the "expired ..., will re-fetch" framing including the boundary `t == now` case.
  - `src/lib/api/__tests__/artifacts.test.ts` — 3 new cases pinning `adaptArtifact` plumbing: fields populated when present, fields `undefined` when omitted, fields `undefined` when the backend explicitly sends `null` (defensive against future serialization changes). Total: 28/28 pass (was 25 before).
- [ ] **E2E Playwright tests added/updated** — Deferred. Realistic E2E coverage requires a Remote-repo fixture with a proxy-cache metadata blob; the unit tests plus the explicit `repo_type === "remote"` gate cover the visibility contract.
- [x] **Manually verified** — `npm run lint` (0 errors, 36 pre-existing warnings unchanged), `npm run build` (clean), `npx tsc --noEmit` (no new errors; same 2 pre-existing on `main` unchanged), `npx vitest run` for both new test files (37/37 pass).
- [x] No regressions in existing tests.

## UI Changes

- [x] Responsive layout verified — both rows reuse the existing `DetailRow` grid (`grid-cols-[100px_1fr]`) used by Created / SHA-256 / Download URL.
- [x] Dark mode verified — uses theme tokens (`text-muted-foreground`); no custom colors.
- [x] Accessibility checked — the visible value is the relative-time string for quick scanning; the hover-tooltip carries the full ISO-8601 timestamp for screen-reader users and operators who need the exact moment. Each row remains a labeled key/value pair, matching the existing dialog rows.
- [ ] N/A - no UI changes

## Companion Work

- artifact-keeper#1541 / [#1542](https://github.com/artifact-keeper/artifact-keeper/pull/1542) — backend exposing `cache_cached_at` and `cache_expires_at` on `GET /:key/artifacts/:path`. **Hard dependency**: the rows render only when the response actually carries the fields. The web change is forward-compatible — until the backend PR lands the rows simply don't appear, no errors.
- Related shipped work in the same dialog: the per-artifact "Invalidate cache" button (#446 / [#447](https://github.com/artifact-keeper/artifact-keeper-web/pull/447)). This PR completes the "see + act" loop alongside it — operators now both *see* the cache state and *act* on it from one panel.


brandonrc (commented):
Accessibility review (WCAG 2.2 AA) — repo-detail-content.tsx Cached / Cache expires DetailRows.

This is display-only and broadly fine: the relative-time strings ("in 4 hours", "expired 12 minutes ago, will re-fetch on next download") are real visible text inside labelled DetailRows, so screen readers read both the label and value. No icon-only controls, no new interactive elements. Two notes:

1. The full ISO timestamp lives only in the `title` attribute (1.1.1 / 1.3.1). `title` is not reliably exposed to screen-reader users and is invisible to keyboard-only users (it only shows on mouse hover). The relative text is the primary, accessible value, so this is acceptable, but if the exact timestamp is considered important info, expose it as visible text or via `aria-label`/visually-hidden text rather than relying on `title` alone. Recommend keeping the relative string as the visible value (good) and treating `title` as purely supplementary.

2. Minor — relative timestamps are computed once at render and go stale (a row reading "in 1 minute" stays that way). Not a WCAG issue, but for the "Cache expires" row consider that a stale "in 1 minute" can mislead; out of scope for a11y sign-off.

No blocking accessibility issues. Approve from an a11y standpoint (leaving formal approval to code owners).

brandonrc (commented):
Frontend correctness review. Clean and well-tested. The `adaptArtifact` runtime-cast approach for the not-yet-in-SDK fields is consistent with the documented apiFetch/SDK-lag pattern, the `?? undefined` normalization (covering both omitted and explicit-null) is correct, and `cache-time.ts` is pure, dependency-free, and deterministically testable via the injected `now`. The `repo_type === "remote"` + truthy-field gates on the detail rows are right. Good unit coverage on the relative-time edge cases.

No blockers. Minor items:

💭 **Missing trailing newline (repo-detail-content.tsx)**
The diff ends with `\ No newline at end of file` on the closing brace of `DetailRow`. If the repo's eslint/prettier enforces `eol-last` this will fail lint. Add the newline.

💭 **`formatRelativeTimestamp` unit-selection loop is correct but the final `return iso` is dead**
The loop always returns on the `unit === "second"` iteration, so the trailing `return iso;` after the loop can never execute. Harmless, but a reader may wonder when it fires. Consider a comment or removing it.

💭 **`formatCacheExpiry` boundary semantics**
Treating `t <= now` as "expired" (including exact equality) is the right product call and you pinned it in a test. No action needed; noting it because it's the kind of thing a future refactor could flip.

Note (not a defect): touches `repo-detail-content.tsx` (shared with #447 and #458/#460/#462/#464) and `src/lib/api/artifacts.ts` + `artifacts.test.ts` (shared with #447 and #454). #447 adds a test block to artifacts.test.ts at a nearby location and #454 edits the same file; expect mechanical conflicts in the test file. Flag merge order. The `adaptArtifact` change here and the `createDownloadTicket` change in #454 are in the same module but different functions, so source-level conflicts are unlikely.

brandonrc (commented):

Reviewed for sensitive-data exposure in the artifact details dialog. The two new rows render `cache_cached_at` and `cache_expires_at` only, formatted through `formatRelativeTimestamp` / `formatCacheExpiry` (timestamps, gracefully falling back to the raw input string for unparseable values). No internal storage keys, filesystem/S3 paths, upstream credentials, or backend object identifiers are surfaced. The rows are gated on `repo_type === "remote"` and on the field being present.

The `adaptArtifact` plumbing surfaces only these two timestamp fields and normalises both missing and explicit-null to `undefined` (tests pin all three cases), so non-remote repos and uncached entries hide the rows rather than rendering empty/"null" text. No new fields beyond the two timestamps ride through the cast.

Rendering is via React JSX (auto-escaped); no `dangerouslySetInnerHTML`. The new `title` prop on `DetailRow` is a string attribute and is also escaped. No XSS surface.

Note: missing trailing newline at EOF on `repo-detail-content.tsx` (cosmetic, lint may flag).

No blocking issues.

brandonrc:
Thank you for this contribution. Your work shipped in **v1.2.0**: the cached_at / expires_at display in the artifact details dialog feature was incorporated into the release integration branch #466, which was squash-merged to main and tagged as [v1.2.0](https://github.com/artifact-keeper/artifact-keeper-web/releases/tag/v1.2.0), resolving #449. The feature is live on main now.

PR #454, opened 2026-06-01
fix(tickets): bind download/stream ticket resource_path to the request path

### LINKED ISSUE #453: Download/migration ticket resource_path does not match request path (UI downloads fail)
## Summary

Downloading an artifact from the web UI, and the migration progress stream, both fail because the download-ticket `resource_path` the web client sends does not match what the backend ticket middleware expects.

Reported by @flopma in artifact-keeper/artifact-keeper#871: downloading a Maven artifact triggers `POST /api/v1/auth/ticket`, which returns:

```
Validation error: resource_path must start with '/' code="VALIDATION..."
```

Originally filed against the backend as artifact-keeper/artifact-keeper#1552, but the backend is behaving as designed. The fix belongs here.

## Root cause

The backend binds a download ticket to an exact request path and, at consume time, compares `bound_path == request.uri().path()` by byte equality (`backend/src/api/middleware/auth.rs`). The minter (this web client) is responsible for sending the exact absolute path the later request will carry. Two call sites send the wrong value:

| File | Sends (wrong) | Must send |
|------|---------------|-----------|
| `src/lib/api/artifacts.ts` (`createDownloadTicket`) | `` `${repoKey}/${artifactPath}` `` | `` `/api/v1/repositories/${repoKey}/download/${artifactPath}` `` |
| `src/lib/api/migration.ts` (`createStreamTicket`) | `` `migration/${jobId}` `` | `` `/api/v1/migrations/${jobId}/stream` `` |

Each is wrong twice: it is not absolute (the immediate `must start with '/'` rejection), and even made absolute it would not byte-match the real request path, so the ticket would be consumed against a mismatching path and silently burned.

## Fix

Bind each ticket to the exact request path the subsequent download/stream uses. Add regression tests asserting the `resource_path` argument passed to the SDK for both call sites.

## Affected versions

Surfaced during the v1.2.0 ticket path-binding work. UI downloads and migration streaming are broken in the 1.2.0 line until this lands.

brandonrc:
## Summary

UI artifact downloads and migration progress streaming are broken because the web client binds download tickets to a `resource_path` the backend cannot match.

The backend binds a download ticket to an exact request path and, at consume time, compares `bound_path == request.uri().path()` by byte equality (`backend/src/api/middleware/auth.rs`). The minter is responsible for sending the exact absolute path the later request will carry. Two call sites sent the wrong value:

| Call site | Sent (wrong) | Now binds |
|-----------|--------------|-----------|
| `artifactsApi.createDownloadTicket` | `` `${repoKey}/${artifactPath}` `` | `` `/api/v1/repositories/${repoKey}/download/${artifactPath}` `` |
| `migrationApi.createStreamTicket` | `` `migration/${jobId}` `` | `` `/api/v1/migrations/${jobId}/stream` `` |

Each was wrong twice: not absolute (the `resource_path must start with '/'` rejection @flopma hit), and not the full request path, so even an absolute form would have been consumed against a mismatching path and silently burned.

No backend change is needed; the backend validator and consumer middleware behave as designed.

Fixes #453

## Test Checklist
- [x] Manually tested locally

Added regression tests asserting the exact `resource_path` passed to the SDK for both call sites (`src/lib/api/__tests__/artifacts.test.ts`, `src/lib/api/__tests__/migration.test.ts`). Ran the affected suites locally: 73 passed. ESLint clean on changed files.

## UI Changes
- [x] N/A - no UI changes

This is a client-side request-payload fix in the API layer. No rendered UI changed, though it restores the download and migration-stream flows the UI depends on.

### DIFF
--- a/src/lib/api/__tests__/artifacts.test.ts
+++ b/src/lib/api/__tests__/artifacts.test.ts
@@ -270,6 +270,26 @@ describe("artifactsApi", () => {
     await expect(artifactsApi.createDownloadTicket("repo-key", "lib.jar")).rejects.toBe("fail");
   });
 
+  it("createDownloadTicket binds resource_path to the absolute download request path", async () => {
+    // The backend ticket middleware compares the bound resource_path against
+    // request.uri().path() by byte equality, so it must be absolute and match
+    // exactly what getDownloadUrl produces. Regression test for web#453.
+    mockCreateDownloadTicket.mockResolvedValue({
+      data: { ticket: "tk123" },
+      error: undefined,
+    });
+    const { artifactsApi } = await import("../artifacts");
+    const artifactPath = "com/example/lib/1.0/lib-1.0.jar";
+    await artifactsApi.createDownloadTicket("repo-key", artifactPath);
+    const expectedPath = artifactsApi.getDownloadUrl("repo-key", artifactPath);
+    expect(expectedPath).toBe(
+      "/api/v1/repositories/repo-key/download/com/example/lib/1.0/lib-1.0.jar"
+    );
+    expect(mockCreateDownloadTicket).toHaveBeenCalledWith({
+      body: { purpose: "download", resource_path: expectedPath },
+    });
+  });
+
   // ---- upload() via XMLHttpRequest ----
 
   describe("upload", () => {
--- a/src/lib/api/__tests__/migration.test.ts
+++ b/src/lib/api/__tests__/migration.test.ts
@@ -582,6 +582,18 @@ describe("migrationApi", () => {
     await expect(migrationApi.createStreamTicket("m1")).rejects.toBe("fail");
   });
 
+  it("createStreamTicket binds resource_path to the absolute stream request path", async () => {
+    // The backend ticket middleware compares the bound resource_path against
+    // request.uri().path() by byte equality, so it must be the absolute path
+    // the EventSource stream request uses. Regression test for web#453.
+    mockCreateDownloadTicket.mockResolvedValue({ data: { ticket: "tk123" }, error: undefined });
+    const { migrationApi } = await import("../migration");
+    await migrationApi.createStreamTicket("m1");
+    expect(mockCreateDownloadTicket).toHaveBeenCalledWith({
+      body: { purpose: "stream", resource_path: "/api/v1/migrations/m1/stream" },
+    });
+  });
+
   it("narrowSourceRepoType warns on unknown type", async () => {
     const warn = vi.spyOn(console, "warn").mockImplementation(() => {});
     mockListSourceRepositories.mockResolvedValue({
--- a/src/lib/api/artifacts.ts
+++ b/src/lib/api/artifacts.ts
@@ -151,8 +151,15 @@ export const artifactsApi = {
   },
 
   createDownloadTicket: async (repoKey: string, artifactPath: string): Promise<string> => {
+    // The backend binds the ticket to this exact path and, at consume time,
+    // compares it against request.uri().path() by byte equality. It must be the
+    // absolute path the subsequent download request will carry, which is the
+    // same value getDownloadUrl produces.
     const { data, error } = await createDownloadTicket({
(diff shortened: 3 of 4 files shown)


brandonrc (commented):
Accessibility review (WCAG 2.2 AA) — download/stream ticket resource_path fix.

This change is confined to API client logic (`createDownloadTicket` / `createStreamTicket` now bind the absolute `resource_path`) plus regression tests. No UI surface, no new controls, labels, dialogs, tables, or status messages. Nothing to assess against WCAG here.

One adjacent observation, not introduced by this PR: ticket-creation failures surface through the existing reject path; ensure wherever these tickets are consumed (download/stream trigger UI) a failed ticket is announced to AT via a live region rather than a silent no-op. Out of scope for this PR.

No accessibility concerns with this change.

brandonrc (commented):
Frontend correctness review. This is the correct fix. I verified the byte-equality contract holds on both sides:

- `createDownloadTicket` now binds `resource_path` to `/api/v1/repositories/${repoKey}/download/${artifactPath}`, which is exactly what `getDownloadUrl` returns (also unencoded). The browser download request in `repo-detail-content.tsx#handleDownload` is `getDownloadUrl(...) + "?ticket=..."`, and the backend compares `request.uri().path()` (path only, query stripped), so the bound path matches the consumed path. Correct.
- `createStreamTicket` now binds `/api/v1/migrations/${jobId}/stream`, which matches the URL `createProgressStream` builds before attaching the ticket as a query param. Correct.

The two regression tests pin the exact request shape, including asserting `getDownloadUrl` produces the same string the ticket binds to, which is the right way to lock this so a future change to one path forces a change to the other.

No blockers, no suggestions of substance.

💭 **Encoding consistency (pre-existing, not introduced here)**
Neither `getDownloadUrl` nor the new bound `resource_path` percent-encode `artifactPath`, so a path with a space or `#` could mismatch what the browser actually sends in `request.uri().path()` (the browser will encode the href). This is a pre-existing concern that predates this PR and the fix deliberately keeps the two sides identical, so it does not regress anything. Worth a follow-up issue to encode consistently on both the URL builder and the ticket binding, but out of scope here.

Touches `src/lib/api/artifacts.ts` / `artifacts.test.ts`, shared with #447 and #451. Different functions, so expect at most a mechanical test-file conflict. Flag merge order.

brandonrc (commented):

Verified against the backend ticket middleware in `backend/src/api/middleware/auth.rs`:

- `ticket_path_allowed` (line 743) does `Some(p) => p == request_path`, an exact byte-equality check.
- `request_path` is `request.uri().path()` (line 859), i.e. the full path including the `/api/v1/...` prefix, with the query string (`?ticket=`) already stripped.

The change is correct. The new `resource_path` (`/api/v1/repositories/${repoKey}/download/${artifactPath}` and `/api/v1/migrations/${jobId}/stream`) now matches exactly what the consuming request carries. The previous values (`${repoKey}/${artifactPath}`, `migration/${jobId}`) could never byte-match `request.uri().path()`, so tickets were effectively always rejected. This is a real fix, not a loosening of the check.

Confirmed the consume path is identical in `repo-detail-content.tsx:182-186`: `getDownloadUrl(repoKey, artifact.path)` builds the href and `createDownloadTicket(repoKey, artifact.path)` mints the ticket from the same raw interpolation, so both stay byte-identical.

No over-broad binding: the bound path is fully-qualified to a single resource, so one ticket authenticates exactly one path (the tightest possible binding). No SSRF/open-redirect: the value is a server-relative path with no host/scheme component, and `repoKey`/`path` originate from the artifact list rather than free-form user input at mint time.

Minor (non-blocking, pre-existing): neither side percent-encodes `artifactPath`, so a path containing `?` or `#` would truncate identically on both sides. Not introduced by this PR and not reachable with valid artifact paths. Consider normalising encoding in `getDownloadUrl` in a follow-up so the mint and consume paths share one helper.

No blocking issues.

brandonrc:

PR #457, opened 2026-06-02
fix: absolute download URL and resilient server version display

### LINKED ISSUE #455: Artifact details 'Download URL' is a host-less path — copy yields a broken URL
## Summary

On the artifact details panel, the **"Download URL"** field is a host-less path, so copy-pasting it produces a broken URL.

## Detail

The field (rendered in `src/app/(app)/repositories/_components/repo-detail-content.tsx`, with a Copy button) takes its value from `artifactsApi.getDownloadUrl` (`src/lib/api/artifacts.ts`):

```ts
getDownloadUrl: (repoKey, artifactPath) =>
  `/api/v1/repositories/${repoKey}/download/${artifactPath}`
```

That's a **relative path** — no scheme or host. Copying the "Download URL" yields e.g.:

```
/api/v1/repositories/my-repo/download/my-repo/1.0.0/image.tar.zst
```

which isn't usable as a URL (a user pasting it into `curl`/a browser gets nothing). First noticed on an `incus` artifact, but the helper is format-agnostic so it affects every format's details panel.

## Expected

A field labelled "Download URL" should be an absolute, copy-pasteable URL, e.g. `https://<ak-host>/api/v1/repositories/.../download/...`.

## Suggested fix

Prepend the app origin (e.g. `window.location.origin`, or the configured API base) when building the value for display/copy — either in `getDownloadUrl` or at the render site. (The path itself is correct; it just needs to be absolute for the copy affordance.)

### LINKED ISSUE #456: Sidebar hides the Server (backend) version when /health is non-2xx, though version is in the body
## Summary

The sidebar shows the **Web** version but usually not the **Server** (backend) version, even though the code is written to show both. The Server version is the one operators care about most, and it disappears precisely when it's most useful.

## Root cause

`app-sidebar.tsx` renders `Web {NEXT_PUBLIC_APP_VERSION}` and, when available, `/ Server {health.version}` from `adminApi.getHealth()` (`GET /health`).

But `/health` returns **HTTP 503** whenever *any* health check is degraded (storage, etc.), and `getHealth()` treats a non-2xx response as a failed query — so the whole response is dropped and `health.version` is `undefined`, leaving only the Web version. Crucially, the 503 **body still contains the version**:

```
GET /health → 503
{"status":"unhealthy","version":"1.2.0-rc.2","checks":{...}}
```

So the Server version is shown only while the backend is fully healthy, and hidden exactly when something is wrong — which is when knowing the running backend version matters most.

## Expected

Show the Server version whenever the backend reports one, independent of overall health status. The web version is comparatively stable; the backend is what's iterated, so its version should be at least as visible.

## Suggested fix

Don't discard the `/health` body on a non-2xx status for the purpose of the version display — parse `version` from the response regardless of status code (or read it from a status-independent source). The health *status* indicator can stay gated on 2xx; only the version string should survive a degraded health response.

brandonrc:
## Summary
Fixes two v1.2.0 web bugs.

**#455** - The artifact detail "Download URL" field rendered a host-less path (e.g. `/api/v1/repositories/.../download/...`), so copying it produced a broken URL. Added `artifactsApi.getAbsoluteDownloadUrl`, which resolves the download path against `NEXT_PUBLIC_API_URL` (falling back to `window.location.origin`), matching how downloads are actually issued. The copyable "Download URL" detail row now uses it. `getDownloadUrl` is unchanged for the same-origin download flow.

**#456** - The sidebar hid the backend (Server) version whenever `/health` returned a non-2xx status, because `adminApi.getHealth` threw on any SDK error. The backend returns 503 with a full `HealthResponse` body (including the version) when a dependency is degraded. `getHealth` now detects the `HealthResponse` shape on the error path and adapts it, so the version stays visible even on a degraded health response.

Closes #455
Closes #456

## Test Checklist
- [ ] Manually tested locally

Unit tests: added `getAbsoluteDownloadUrl` coverage in `artifacts.test.ts` and a degraded-503 case in `admin.test.ts`. E2E: new spec `e2e/suites/interactions/repositories/download-url-and-version.spec.ts` asserts the copied Download URL is absolute and parseable, and that the sidebar shows the backend version (cross-checked against the live `/health` body). Ran eslint, tsc, and vitest on the changed files; all green.

## UI Changes
- [ ] Dark mode verified

The change is to the value of an existing copyable field and an existing sidebar version label; no layout, styling, or markup structure changed.

### DIFF
--- a/e2e/suites/interactions/repositories/download-url-and-version.spec.ts
+++ b/e2e/suites/interactions/repositories/download-url-and-version.spec.ts
@@ -1,0 +1,123 @@
+import { test, expect } from '@playwright/test';
+
+/**
+ * Regression tests for two v1.2.0 web bugs.
+ *
+ * #455 — The artifact detail "Download URL" field showed a host-less path
+ *        (e.g. /api/v1/repositories/.../download/...), so copying it produced a
+ *        broken URL. The displayed value must be a full, absolute URL.
+ *
+ * #456 — The sidebar hid the backend (Server) version whenever /health returned
+ *        a non-2xx status, even though the version is present in the response
+ *        body. The version must be shown regardless of the /health status code.
+ */
+test.describe('Download URL and version display', () => {
+  const REPO_KEY = 'e2e-maven-local';
+  const ARTIFACT_CONTENT = 'Hello from Playwright download-url test';
+
+  async function uploadArtifact(
+    request: import('@playwright/test').APIRequestContext
+  ): Promise<string> {
+    const path = `e2e/download-url-test-${Date.now()}.txt`;
+    const resp = await request.put(
+      `/api/v1/repositories/${REPO_KEY}/artifacts/${path}`,
+      {
+        data: ARTIFACT_CONTENT,
+        headers: { 'Content-Type': 'application/octet-stream' },
+      }
+    );
+    expect(resp.ok(), `Upload failed: ${resp.status()}`).toBeTruthy();
+    const body = await resp.json();
+    return body.path || path;
+  }
+
+  test('Download URL in detail panel is an absolute, well-formed URL (#455)', async ({
+    page,
+    request,
+  }) => {
+    const artifactPath = await uploadArtifact(request);
+    const artifactName = artifactPath.split('/').pop()!;
+
+    await page.goto(`/repositories/${REPO_KEY}`);
+    await page.waitForLoadState('domcontentloaded');
+
+    const table = page.getByRole('table').first();
+    await expect(table).toBeVisible({ timeout: 15000 });
+
+    const searchInput = page.getByPlaceholder(/search/i).first();
+    if (await searchInput.isVisible({ timeout: 3000 }).catch(() => false)) {
+      await searchInput.fill(artifactName);
+      await page.waitForTimeout(2000);
+    }
+
+    const artifactRow = table.getByRole('row').filter({ hasText: artifactName });
+    if (!(await artifactRow.isVisible({ timeout: 5000 }).catch(() => false))) {
+      await request
+        .delete(`/api/v1/repositories/${REPO_KEY}/artifacts/${artifactPath}`)
+        .catch(() => {});
+      test.skip(true, 'Artifact not visible in table');
+      return;
+    }
+
+    await artifactRow.click();
+    await page.waitForTimeout(1500);
+
+    const downloadUrlLabel = page.getByText('Download URL');
+    if (!(await downloadUrlLabel.isVisible({ timeout: 5000 }).catch(() => false))) {
+      await request
+        .delete(`/api/v1/repositories/${REPO_KEY}/artifacts/${artifactPath}`)
+        .catch(() => {});
+      test.skip(true, 'Download URL field not visible in detail panel');
+      return;
+    }
+
+    // The value lives in the detail row next to the "Download URL" label.
+    const urlContainer = downloadUrlLabel.locator('..').locator('..');
+    const urlText = (await urlContainer.textContent())?.trim() ?? '';
+
+    // Extract the URL portion (the row also contains the "Download URL" label).
+    const match = urlText.match(/https?:\/\/\S+/);
+    expect(
+      match,
+      `Download URL should be absolute (http/https). Got: "${urlText}"`
+    ).not.toBeNull();
+
+    const url = match![0];
+
+    // Must be parseable and carry an origin + the backend download route.
+    let parsed: URL | null = null;
+    expect(() => {
+      parsed = new URL(url);
+    }).not.toThrow();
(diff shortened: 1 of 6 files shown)


brandonrc (commented):
Test/E2E quality review (v1.2.0 batch). Verdict: real coverage, with one weak spot.

**#455 download URL.** The unit test `getAbsoluteDownloadUrl returns a full, well-formed URL` is the load-bearing assertion and it is honest: it stubs `NEXT_PUBLIC_API_URL` and asserts the exact absolute string plus `new URL(url).protocol === "https:"`. The pre-fix `getDownloadUrl` returned a host-less `/api/v1/...` path, which would fail both `new URL(url)` (throws without a base) and the `protocol` check, so this assertion genuinely distinguishes fixed-vs-broken. Good.

**#455 e2e.** This one is softer than it looks. There are two `test.skip(...)` escape hatches (artifact not in table, Download URL label not visible). If either fires, the bug-fix assertion (`/https?:\/\/\S+/` regex + `new URL`) never runs and the test still goes green. The skip on "Download URL field not visible" is the concerning one: a regression that broke the detail panel rendering would skip rather than fail. Since the unit test already nails the actual fix, this is acceptable, but consider asserting the row is present (hard fail) and only skipping on the upload/seed step.

**#456 version.** Strong. The unit test `getHealth uses the body version on a non-2xx degraded response` mocks `{ data: undefined, error: degraded }` and asserts `result.version === "1.2.0"`. Pre-fix `getHealth` did `if (error) throw error`, so this test would throw and fail. It directly proves the fix. The e2e grounds the assertion in the live `/health` body (`expect(text).toContain('Server ${expectedVersion}')`) only when a version is present, which is the right call.

No flaky arbitrary-ordering issues. Net: genuine coverage. Not requesting changes.

brandonrc (commented):
Security review (open redirect / host injection / data trust lens).

getAbsoluteDownloadUrl — SAFE, no open redirect or host injection. The origin comes only from `NEXT_PUBLIC_API_URL` (a build-time operator-controlled env var) or `window.location.origin` (the browser's own trusted origin), never from user input or the artifact record. `new URL(path, origin)` resolves a fixed local path against that trusted origin, so an attacker who controls `artifactPath` cannot redirect to a foreign host: the path is the second-position relative ref and the origin is the base, so even an absolute-looking `artifactPath` stays under `/api/v1/repositories/.../download/...`. The value is rendered into a copyable text field (`DetailRow ... copy mono`), not used as a live `href`/redirect target, so there is no navigation sink. No SSRF either since this is client-side display only.

One note, not a defect: if `artifactPath` ever contained a `?` or `#`, `new URL` would split it into query/fragment, but the copied string remains same-origin and harmless; backend routing already governs the real download.

getHealth trusting a non-2xx body — ACCEPTABLE. On a 503 the SDK surfaces the parsed body as `error`; `isSdkHealthResponse` narrows it (`version` is string, `checks` is a non-null object) before `adaptHealth`. The body is the backend's own /health payload over the same trusted channel, and the consumed fields are just version/status/check strings rendered as text in the sidebar. No injection sink, no privilege decision keyed off this. Worst case a malformed-but-shaped body shows a wrong version string. Low risk.

No request-changes. Looks good to ship.

brandonrc (commented):
Frontend/React correctness review. No blockers. Both fixes are correct and well-tested.

### Correctness notes

- `getAbsoluteDownloadUrl` resolves against `NEXT_PUBLIC_API_URL` then `window.location.origin`, with a `new URL(...)` try/catch fallback. Solid. One edge: if `NEXT_PUBLIC_API_URL` carries a path prefix (e.g. `https://host/registry`), `new URL("/api/v1/...", "https://host/registry")` resolves against the origin and drops `/registry`. The catch-branch concatenation (`origin + path`) would keep it, but it only runs on a thrown error, which `new URL` won't throw here. Likely a non-issue for current deployments (the SDK base is typically bare origin), just flagging.

- Multi-instance gap: `sdk-client.ts` exposes `getActiveInstanceBaseUrl()`, which prefixes `/api/v1/instances/{id}/proxy` when a non-local instance is active. `getAbsoluteDownloadUrl` only uses `NEXT_PUBLIC_API_URL`/origin, so when viewing an artifact through a proxied remote instance the copied URL points at the local base rather than the proxied path. The same-origin `getDownloadUrl` download flow relies on the Next rewrite and is unaffected. If artifact detail can be opened under an active remote instance, the copied URL would be wrong there. Worth a follow-up; not blocking the primary same-origin case.

- The degraded-health adapter (`isSdkHealthResponse` + adapt-on-error in `adminApi.getHealth`) is a clean, well-scoped fix: it only treats the error body as a health response when it structurally looks like one (`version` string + `checks` object), otherwise rethrows. Good. The unit test covers the 503-with-body path.

Nice, focused PR.

brandonrc:

PR #458, opened 2026-06-02
feat(maven): search by GAV and surface POM in the UI

### LINKED ISSUE #441: Cannot search Maven artifacts from its GAV information
### Problem Statement

For Maven repositories, one important way to search for artifacts is by looking up with any number of these different informations
- groupId
- artifactId
- version
- classifier
- file extension

It is currently impossible to perform such a search in v1.2.0

### Proposed Solution

Add a search functionality for Maven repository which allows to perform searches with any of these information filled
- groupId
- artifactId
- version
- classifier
- file extension

### Alternatives Considered

_No response_

### Use Case

_No response_

### Component

Search

### Additional Context

_No response_

### Pre-submission Checklist

### LINKED ISSUE #442: Can't find POM file of GAV data for a Maven artifact
### Problem Statement

AFAIK, in the repository browser, you do not find the POM file listed, and cannot download it by browsing the repository.
When opening the a Maven artifact popup with details, you also do not find the GAV displayed

### Proposed Solution

List the related POM file of a Maven artifact, since it is itself a artifact of type pom

Also, display the GAV data for each Maven artifacts and possibly a snippet to copy/paste into a pom.xml file

### Alternatives Considered

_No response_

### Use Case

_No response_

### Component

Package Details

### Additional Context

_No response_

### Pre-submission Checklist

brandonrc:
## Summary

Two Maven UX fixes for v1.2.0.

**GAV search (Closes #441).** The Advanced Search GAVC tab sent groupId, artifactId, and version as the backend's `path`/`name`/`version` query params. The advanced-search endpoint ignores those: it matches a single full-text `query` against name + path + version. Because Maven coordinates live in the artifact path, GAVC searches returned nothing. The tab now folds the GAV fields, classifier, and a new **Extension** field into one full-text query scoped to the `maven` format. A shared `buildMavenSearchQuery` helper assembles the terms.

**POM and GAV display (Closes #442).** In the grouped Maven browser, each component file (jar, pom, checksums) now renders as a download link with the path derived from the GAV layout, and the `.pom` file is tagged with a POM badge so it is easy to find and download. The artifact detail view gains a Maven section that shows the parsed groupId / artifactId / version and a copy/paste `pom.xml` dependency snippet.

New `src/lib/maven.ts` holds the pure coordinate helpers (query building, GAV-to-path, POM detection, GAV parsing, dependency snippet).

## Test Checklist
- [x] Manually tested locally

## UI Changes
- [x] Dark mode verified
- [ ] N/A - no UI changes

Validation: `eslint` clean on changed files, `tsc --noEmit` reports only two pre-existing errors in untouched test files, 16 new unit tests pass, existing maven-component-list (16) and search-content (48) suites still pass.

### DIFF
--- a/e2e/suites/interactions/repositories/maven-gav-search-pom.spec.ts
+++ b/e2e/suites/interactions/repositories/maven-gav-search-pom.spec.ts
@@ -1,0 +1,152 @@
+import { test, expect } from '@playwright/test';
+
+/**
+ * Regression tests for Maven GAV search (issue #441) and POM reachability
+ * (issue #442).
+ *
+ * - #441: users must be able to find a Maven artifact from its GAV coordinates
+ *   (groupId / artifactId / version / classifier / extension) via the GAVC tab
+ *   on the Advanced Search page.
+ * - #442: the POM file of a Maven GAV must be listed and downloadable from the
+ *   repository browser, and the artifact detail view must surface the GAV.
+ *
+ * The fix routes GAVC fields into the backend's full-text `query` (the
+ * advanced-search endpoint matches name + path + version, it does not filter on
+ * the separate path/version params) and renders each component file as a
+ * download link plus a GAV section in the detail panel.
+ */
+test.describe('Maven GAV search and POM', () => {
+  const REPO_KEY = 'e2e-maven-local';
+  // Unique per run so parallel/retried runs do not collide.
+  const STAMP = Date.now();
+  const GROUP_ID = `com.akweb.e2e${STAMP}`;
+  const GROUP_PATH = GROUP_ID.replace(/\./g, '/');
+  const ARTIFACT_ID = 'gav-search-lib';
+  const VERSION = '1.0.0';
+  const BASE = `${GROUP_PATH}/${ARTIFACT_ID}/${VERSION}`;
+  const JAR_NAME = `${ARTIFACT_ID}-${VERSION}.jar`;
+  const POM_NAME = `${ARTIFACT_ID}-${VERSION}.pom`;
+  const JAR_PATH = `${BASE}/${JAR_NAME}`;
+  const POM_PATH = `${BASE}/${POM_NAME}`;
+
+  const POM_CONTENT = [
+    '<?xml version="1.0" encoding="UTF-8"?>',
+    '<project xmlns="http://maven.apache.org/POM/4.0.0">',
+    '  <modelVersion>4.0.0</modelVersion>',
+    `  <groupId>${GROUP_ID}</groupId>`,
+    `  <artifactId>${ARTIFACT_ID}</artifactId>`,
+    `  <version>${VERSION}</version>`,
+    '</project>',
+  ].join('\n');
+
+  async function put(
+    request: import('@playwright/test').APIRequestContext,
+    path: string,
+    data: string,
+    contentType: string
+  ): Promise<void> {
+    const resp = await request.put(
+      `/api/v1/repositories/${REPO_KEY}/artifacts/${path}`,
+      { data, headers: { 'Content-Type': contentType } }
+    );
+    expect(resp.ok(), `Upload of ${path} failed: ${resp.status()}`).toBeTruthy();
+  }
+
+  test.beforeAll(async ({ request }) => {
+    await put(request, JAR_PATH, 'fake-jar-bytes-for-gav-search', 'application/java-archive');
+    await put(request, POM_PATH, POM_CONTENT, 'application/xml');
+  });
+
+  test.afterAll(async ({ request }) => {
+    await request.delete(`/api/v1/repositories/${REPO_KEY}/artifacts/${JAR_PATH}`).catch(() => {});
+    await request.delete(`/api/v1/repositories/${REPO_KEY}/artifacts/${POM_PATH}`).catch(() => {});
+  });
+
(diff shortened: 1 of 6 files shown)


brandonrc (commented):
Test/E2E quality review (v1.2.0 batch). Verdict: strongest unit coverage in the batch; e2e has soft spots.

**maven.test.ts (#441/#442).** Excellent. `buildMavenSearchQuery` is covered for join, empty/whitespace skipping, trimming, leading-dot strip, and empty-input. `parseMavenGav` covers standard layout, leading slash, and the negative case (non-Maven paths return undefined). `isPomFile` covers the `.pom.sha1`/`.pom.asc` false-positive traps, which is exactly the kind of negative case that proves the fix is real. These tests would fail against the old code path (which fed `path`/`name`/`version` separately instead of folding into `query`).

**#442 POM download e2e.** Good and deterministic: PUTs the POM via API, then `GET /download/${POM_PATH}` asserts status 200 and body contains the artifactId. This is a pure contract assertion with no skip path.

**#441 GAVC search e2e.** Real: fills group/artifact/version/extension, clicks search, asserts the JAR row appears in results. No skip escape. The `#search-gavc-extension` visibility assertion ties directly to the fix. Reasonable.

**#442 grouped-browser + detail e2e.** Three `test.skip` paths here (GAV row not rendered, flat table not rendered, jar not visible). The detail-view test for `maven-gav-section` / `maven-pom-snippet` is the assertion that proves #442's GAV-surfacing fix, and it sits behind two skips. If the grouped view or flat table regresses, these skip silently green. The unit tests cover the helpers, but the UI-rendering of the GAV section is only exercised by this skippable e2e. Consider a vitest render test for `MavenGavSection` so the snippet rendering has a non-skippable guard.

Minor flake risk: `page.waitForTimeout(2000)` after the search fill is an arbitrary wait; prefer waiting on the row/response. Net: solid, not requesting changes.

brandonrc (commented):
Security/API review (path traversal + injection lens on Maven path construction).

Net: safe. No traversal sink and no HTML/XSS injection introduced.

mavenFilePath / parseMavenGav / groupIdToPath (src/lib/maven.ts) — SAFE. These build a repository-relative display/download path from component fields and parse coordinates back out. They are pure string transforms; the result is handed to `artifactsApi.getDownloadUrl` (which produces a same-origin `/api/v1/repositories/<key>/download/<path>` URL rendered as an `href`) or used as display text. A crafted GAV cannot escape the app origin because the path is always appended under the fixed `/download/` route on the current origin; there is no filesystem access on the client. `..` segments would simply be sent to the backend, which is the authoritative traversal-prevention boundary for download/metadata routes (out of scope for this web change, but worth confirming the backend rejects `..` in the wildcard path — see note on #460).

pom snippet rendering — SAFE. `buildPomDependencySnippet` interpolates groupId/artifactId/version into an XML-looking string, but `MavenGavSection` renders it inside `<pre data-testid="maven-pom-snippet">{snippet}</pre>` as a React text child, so it is HTML-escaped, not parsed. No `dangerouslySetInnerHTML`. The copy button copies the raw string. No injection sink.

Encoded separators — see my note on #460: this PR's links go through `getDownloadUrl`, which on main does not per-segment-encode; #460 fixes that for the metadata `get`. Make sure #460 lands with/before this so generated links with spaces or reserved chars resolve correctly. Not a security defect, a correctness coupling.

No request-changes.

brandonrc (commented):
Frontend/React correctness review. One build-breaking blocker, otherwise the Maven helper logic is clean and well-tested. (Posting as a comment since the review tool refuses request-changes on a self-authored PR, but please treat the first item as a blocker.)

### Blocker

**Undefined `repoFormat` in `repo-detail-content.tsx`**
The new Maven GAV section is gated on a `repoFormat` identifier that does not exist in this file:

```tsx
{(repoFormat === "maven" || repoFormat === "gradle") && (
  <MavenGavSection path={selectedArtifact.path} />
)}
```

Everywhere else the file uses `repository.format` (around lines 440, 482, 590); there is no `repoFormat` in scope and the diff does not introduce one. This will not type-check, and would be a `ReferenceError` at runtime. Change to `repository.format === "maven" || repository.format === "gradle"` (or destructure a local `const repoFormat = repository.format`). The PR notes claim `tsc --noEmit` reports only two pre-existing errors, which is inconsistent with this usage, so please re-run `tsc` on the branch and confirm.

### Suggestions

- `src/lib/maven.ts` `parseMavenGav`: the "groupId = all segments before the last three" heuristic is reasonable and the `if (!gav) return null` guard in `MavenGavSection` keeps a misparse harmless (section just hides). No change needed, flagging the limitation for snapshot/timestamped paths.
- File-row links correctly use `artifactsApi.getDownloadUrl` (relative path) for `<a download>` on the same origin. This is now intentionally distinct from #457's `getAbsoluteDownloadUrl` (relative for issuing downloads, absolute for copy-paste). Fine.

### Merge sequencing (not a code defect)

This PR and #463 both rewrite the `gavc` branch of `buildSearchParams` in `search-content.tsx`. #463 reverts that branch to sending `path`/`name`/`version` (the very params this PR shows the backend ignores). Whichever merges second must preserve this PR's `buildMavenSearchQuery` + `format: "maven"` mapping for the gavc tab, or GAV search regresses. Please coordinate the rebase.

Tests are thorough. Once `repoFormat` is fixed this is good.

brandonrc:

PR #459, opened 2026-06-02
fix(a11y): repo dialog and SSO form WCAG 2.2 AA gaps

### LINKED ISSUE #410: a11y: repo edit-dialog upstream-auth save has no screen-reader feedback
## Summary
The "Save Authentication" button in the repository **edit** dialog (`src/app/(app)/repositories/_components/repo-dialogs.tsx` ~L783-807) calls `onUpstreamAuthUpdate` and shows no toast, no live-region announcement, and no visible confirmation. Screen-reader users have no signal that the save succeeded or failed.

This is the *edit* path; the *create* path is fine (dialog close + list update is perceivable). Surfaced during the #407 a11y pre-audit (finding S1) but deferred because the #407 fix only touched the API layer, not this UI path.

## WCAG criterion
4.1.3 Status Messages (Level AA).

## Repro
1. Open a remote repository → Edit.
2. In the "Upstream Authentication" section, click "Configure" or "Change".
3. Enter credentials, click "Save Authentication".
4. **Expected:** SR announces "Upstream authentication saved" (or failure).
5. **Actual:** silence. The view-mode panel re-renders, but with no focus change or live region.

## Suggested fix
- Repo-dialogs.tsx does NOT currently import `toast` from `sonner`. The parent component (`repositories-content.tsx`) does the mutation invocations and could pipe success/error into a `toast.success("Upstream authentication saved")` / `toast.error(...)`. Sonner already provides a polite live region globally, so a single `toast.*` call is sufficient.
- Alternatively add a `role="status"` `aria-live="polite"` element inside the edit dialog that flips to the success/error string when `upstreamAuthPending` transitions back to false.

## Out of scope
The error-summary / aria-describedby story for individual fields is tracked separately (see follow-up #2 below).

## Context
Discovered during PR for #407 / #406 / #378. Out of scope of that PR per minimum-viable-diff guidance.

### LINKED ISSUE #411: a11y: repo create-dialog form errors not linked via aria-describedby + role=alert
## Summary
Inline form errors in the repository **create** dialog (`src/app/(app)/repositories/_components/repo-dialogs.tsx`) are rendered as styled `<p>` elements with no `id`, no `role="alert"`, and no `aria-describedby` linkage on the related input. The submit button is disabled when the form is invalid, but screen-reader users have no path to understand *why*.

Concrete case: the "Repository key already taken" message at ~L256-260.

This is the create-dialog version of the same pattern. The codebase has zero usage of `aria-describedby`, `aria-invalid`, or `role="alert"` across the three forms touched in PR for #407/#406/#378 — so a systematic pass on this file would help.

## WCAG criteria
- 3.3.1 Error Identification (Level A)
- 4.1.3 Status Messages (Level AA)

## Fields to address (minimum)
| Field | File:line | Current state | Needed |
|---|---|---|---|
| Repository Key (taken) | repo-dialogs.tsx:256-260 | `<p>` only, button disabled silently | `<p id="create-key-error" role="alert">`; Input gets `aria-describedby="create-key-error"` and `aria-invalid={keyTaken}` |
| Upstream URL helper | repo-dialogs.tsx:361-363 | `<p>` only | Add id; Input gets `aria-describedby` |
| Upstream auth helper | repo-dialogs.tsx:381-383 | `<p>` only | Add id; relevant Inputs get `aria-describedby` |

## Suggested fix
Generic helper pattern: every `space-y-2` wrapper that contains Label + Input + helper/error `<p>` should give the `<p>` a deterministic id (e.g. `${inputId}-help` / `${inputId}-error`), and the Input should pick that up via `aria-describedby`. `aria-invalid` flips on error. Errors get `role="alert"` so they're announced immediately.

## Out of scope
- Edit-dialog auth save announcement is tracked separately at #410.
- Edit-dialog focus management for view↔edit toggle is tracked separately (next issue).

## Context
Discovered during a11y pre-audit for PR #407 / #406 / #378. Pre-existing, out of scope of that PR.

### LINKED ISSUE #412: a11y: repo edit-dialog auth view↔edit toggle does not manage focus
## Summary
In the repository **edit** dialog's "Upstream Authentication" panel (`src/app/(app)/repositories/_components/repo-dialogs.tsx` ~L642-716), the user toggles between a "view" mode (showing "Authentication configured (Basic Auth)" + Change/Remove buttons) and an "edit" mode (showing the auth-type Select + username/password inputs).

When the user clicks "Configure" or "Change" (L681 / L654), `editAuthMode` flips to `"edit"` and new form inputs render — but keyboard focus stays on the now-removed button. SR users land in dead-air; sighted keyboard users land on the dialog body and have to tab through the entire dialog again to reach the new fields.

## WCAG criterion
2.4.3 Focus Order (Level A). Also 3.2.2 On Input by extension — the UI changes in a way the user can't anticipate the focus destination of.

## Repro
1. Edit a remote repo that has upstream auth already configured.
2. Tab focus to "Change" button. Press Enter.
3. **Expected:** focus moves into the first new input (auth-type Select or username field).
4. **Actual:** focus stays on the now-invisible "Change" button (or jumps to dialog body).

Repeat with "Configure" on a repo without auth — same problem.

## Suggested fix
Wrap the edit-mode inputs in a ref, and in a `useEffect` watching `editAuthMode`, focus the first input when it transitions to `"edit"`. The Cancel button at L770-782 should mirror the inverse (return focus to "Change"/"Configure" when transitioning back to `"view"`).

```tsx
const firstAuthFieldRef = useRef<HTMLButtonElement>(null);
useEffect(() => {
  if (editAuthMode === "edit") firstAuthFieldRef.current?.focus();
}, [editAuthMode]);
```

## Out of scope
- Save announcement on success is tracked at #410.
- Error/describedby linkage is tracked at #411.

## Context
Discovered during a11y pre-audit for PR #407 / #406 / #378. Pre-existing, out of scope of that PR.

### LINKED ISSUE #413: a11y: WCAG 2.2 AA gap audit — repo dialogs + SSO settings forms
## Summary
Batch issue covering lower-severity findings from the a11y pre-audit conducted during PR #407 / #406 / #378. Three high-impact items got separate issues (#410 silent edit-auth save, #411 form-error linking, #412 view↔edit focus management); this issue tracks the remaining S3/S4/S5/S7/S8/S9/S10/S12 findings so they can be batched.

### `src/app/(app)/repositories/_components/repo-dialogs.tsx`

**S3 — Helper-text not linked (Moderate, WCAG 1.3.1)**
- L361-363 (upstream URL helper): `<p>` not linked to `#create-upstream` via `aria-describedby`.
- L381-383 (upstream auth helper): `<p>` not linked to the auth inputs.
Fix: give each `<p>` an `id`, add `aria-describedby` on the related Input.

**S4 — Upstream Authentication grouping uses bare `<Label>` without `htmlFor` (Moderate, WCAG 1.3.1 / 4.1.2)**
- Create dialog L368-424 and Edit dialog L634-812 both wrap multiple inputs under one `<Label>` (no `htmlFor`).
- L637 has a bare `<Label>` with no htmlFor at all.
Fix: replace the wrapping div + bare Label with `<fieldset><legend>Upstream Authentication</legend>…</fieldset>` so SR users perceive the group.

**S5 — Format/Type Selects missing label association (Moderate, WCAG 1.3.1)**
- L286-340: `<Label>Format</Label>` / `<Label>Type</Label>` have no `htmlFor`; SelectTrigger has no `id`. Falls back to whatever the shadcn Select internally does, which is insufficient.
Fix: add `id` to SelectTrigger + `htmlFor` to Label, or `aria-labelledby` pointing at a Label id.

**S7 — Password inputs lack show/hide toggle (Minor, usability)**
- L397-405, L412-420, L742-750, L757-765 are bare `<Input type="password">`.
Fix (optional): add a `<button type="button" aria-pressed={visible} aria-label={visible ? "Hide password" : "Show password"}>` toggle. Not a strict WCAG fail but a strong recommended pattern.

**S8 — Submit buttons change visible text only; not announced after the fact (Minor, WCAG 4.1.3)**
- L513-515, L617-619, L822-824: button text flips "Create" → "Creating…" / "Save" → "Saving…" but a SR user past the button won't perceive the change.
Fix: `aria-busy={pending}` on the form, or an `aria-live="polite"` wrapper around the button text.

### `src/app/(app)/(admin)/settings/sso/page.tsx`

**S9 — "Attribute Mapping" group is a styled `<p>`, not a heading or fieldset (Moderate, WCAG 1.3.1)**
- L426: bare `<p className="text-sm font-medium mb-3">Attribute Mapping</p>` then 5 inputs below it.
Fix: convert to `<fieldset><legend className="text-sm font-medium mb-3">Attribute Mapping</legend>…</fieldset>`. Related a11y-adjacent note from PR #407/#406/#378: the form silently round-trips backend-managed `attribute_mapping` keys (fix in #406) the user cannot see — operators have no signal that other keys exist. Worth surfacing on this form too.

**S10 — Client Secret "Leave blank to keep existing" cue is placeholder-only (Moderate, WCAG 3.3.2)**
- L390-392: critical instruction for edit mode is in `placeholder=`, which disappears on focus and isn't reliably announced.
Fix: move to `<p id="oidc-client-secret-help">` and reference via `aria-describedby` on the Input. Show on edit mode only.

**S12 — Loader spinner has no accessible label (Minor, WCAG 1.1.1)**
- L488: `<Loader2 className="size-4 animate-spin mr-1.5" />` followed by visible "Saving" text.
Fix: `<Loader2 aria-hidden="true" />` + `<span className="sr-only">Saving</span>` if the visible text isn't already serving as the label (it is here, so just `aria-hidden` on the icon).

## Suggested approach
This is a single PR's worth of work if batched. Recommend dispatching one a11y-focused PR rather than threading these through unrelated feature work. The fixes are systematic (label association patterns, fieldset/legend usage, aria-describedby wiring) and will surface a reusable pattern for the codebase.

## Context
Discovered during a11y pre-audit for PR #407 / #406 / #378. All pre-existing; not regressed by that PR.

brandonrc:
## Summary

Fixes four WCAG 2.2 AA accessibility gaps in the repository create/edit dialogs and the SSO settings forms (v1.2.0).

- **#410** Repo edit dialog now mirrors the upstream-auth save outcome into an in-dialog aria-live region. Success renders `role="status"` (`aria-live="polite"`); failure renders `role="alert"` (`aria-live="assertive"`). Previously the only feedback was a visual toast, which assistive tech outside the dialog did not reliably announce while focus stayed inside the open dialog. Wired from the parent mutation in `repositories-content.tsx`.
- **#411** Create dialog "key already taken" error now has `role="alert"`, and the key input sets `aria-invalid` and `aria-describedby` pointing at the message id, so screen readers announce the error and tie it to the field.
- **#412** Toggling the upstream-auth view to edit (and back) now moves focus to the first control of the newly-revealed view (the labelled auth-type select in edit mode, the Change/Configure button in view mode) instead of dropping focus on `<body>`. Implemented with an effect that targets the control by id and defers a frame so the new control exists in the DOM.
- **#413** Broader gap fixes: required inputs across the repo dialogs and SSO OIDC/LDAP/SAML forms expose `aria-required`; the edit-dialog auth-type select gains a programmatic `<label>`; the repository row actions menu trigger gains an accessible name; the edit-key warning is associated via `aria-describedby`.

Closes #410
Closes #411
Closes #412
Closes #413

## Test Checklist
- [x] Manually tested locally

New spec `e2e/suites/interactions/repositories/repo-dialog-a11y.spec.ts` asserts the error/alert + `aria-describedby` association (#411), focus movement on the auth toggle and the auth-type select label (#412/#413), and the live-region save announcement (#410). Existing unit suites (`repo-dialogs.test.tsx` 77 tests, repositories `page.test.tsx`, SSO `page.test.tsx`) pass unchanged. `eslint` clean on changed files (one pre-existing unrelated warning); `tsc --noEmit` reports only pre-existing errors in untouched test files.

## UI Changes
- [ ] Dark mode verified
- [ ] N/A - no UI changes

No visual layout changes; additions are ARIA attributes, a focus-management effect, and a status text line in the upstream-auth section of the edit dialog.

### DIFF
--- a/e2e/suites/interactions/repositories/repo-dialog-a11y.spec.ts
+++ b/e2e/suites/interactions/repositories/repo-dialog-a11y.spec.ts
@@ -1,0 +1,161 @@
+import { test, expect } from '@playwright/test';
+
+/**
+ * Accessibility regression tests for the repository create/edit dialogs
+ * (issues #410, #411, #412, #413). These assert the concrete WCAG 2.2 AA
+ * behaviours that automated scanners miss:
+ *
+ *  - #411: the create-dialog "key already taken" error has role="alert" and
+ *          is referenced by the input's aria-describedby (and aria-invalid).
+ *  - #412: toggling the upstream-auth view <-> edit moves focus to the first
+ *          control of the newly-revealed view (no focus dumped on <body>).
+ *  - #410: saving upstream auth announces the outcome via an in-dialog live
+ *          region (role="status" / aria-live).
+ *  - #413: required inputs expose aria-required; the edit auth-type select has
+ *          a programmatic label.
+ *
+ * Runs against the 1.2.0 backend image. The suite is serial so the seeded
+ * repos are created, exercised, and torn down in order.
+ */
+test.describe.serial('Repository Dialog Accessibility', () => {
+  const REMOTE_KEY = 'e2e-a11y-remote';
+  const EXISTING_KEY = 'e2e-a11y-existing';
+
+  test.beforeAll(async ({ request }) => {
+    // A remote repo whose edit dialog exposes the upstream-auth section.
+    await request.post('/api/v1/repositories', {
+      data: {
+        key: REMOTE_KEY,
+        name: 'E2E A11y Remote',
+        description: 'a11y test fixture',
+        format: 'npm',
+        repo_type: 'remote',
+        is_public: true,
+        upstream_url: 'https://registry.npmjs.org',
+      },
+    }).catch(() => {});
+
+    // A second repo whose key we will collide with in the create dialog.
+    await request.post('/api/v1/repositories', {
+      data: {
+        key: EXISTING_KEY,
+        name: 'E2E A11y Existing',
+        format: 'generic',
+        repo_type: 'local',
+        is_public: true,
+      },
+    }).catch(() => {});
+  });
+
+  test.afterAll(async ({ request }) => {
+    await request.delete(`/api/v1/repositories/${REMOTE_KEY}`).catch(() => {});
+    await request.delete(`/api/v1/repositories/${EXISTING_KEY}`).catch(() => {});
+  });
+
+  test('#411 create dialog duplicate-key error is role=alert and linked via aria-describedby', async ({ page }) => {
+    await page.goto('/repositories');
+    await page.waitForLoadState('domcontentloaded');
+
+    const createBtn = page.getByRole('button', { name: /create repository/i }).first();
+    await expect(createBtn).toBeVisible({ timeout: 10000 });
+    await createBtn.click();
+
+    const dialog = page.getByRole('dialog');
+    await expect(dialog).toBeVisible({ timeout: 5000 });
+
+    const keyInput = dialog.locator('#create-key');
+    await expect(keyInput).toBeVisible({ timeout: 5000 });
+
+    // Type the key of an existing repo to trigger the "already taken" state.
+    await keyInput.fill(EXISTING_KEY);
+
+    // The error message must be an alert so screen readers announce it.
+    const alert = dialog.getByRole('alert');
+    await expect(alert).toBeVisible({ timeout: 5000 });
+    await expect(alert).toContainText(/already taken/i);
+
+    // The input must mark itself invalid and point at the alert.
+    await expect(keyInput).toHaveAttribute('aria-invalid', 'true');
+    const describedBy = await keyInput.getAttribute('aria-describedby');
+    expect(describedBy, 'key input should reference the error via aria-describedby').toBeTruthy();
+    const errorId = await alert.getAttribute('id');
+    expect(describedBy).toBe(errorId);
+
+    // #413: required inputs expose aria-required.
+    await expect(dialog.locator('#create-name')).toHaveAttribute('aria-required', 'true');
+
+    await dialog.getByRole('button', { name: /cancel/i }).click();
+    await expect(dialog).not.toBeVisible({ timeout: 5000 });
+  });
+
+  async function openRemoteEditDialog(page: import('@playwright/test').Page) {
+    await page.goto(`/repositories?selected=${REMOTE_KEY}`);
+    await page.waitForLoadState('domcontentloaded');
+    await page.waitForTimeout(2000);
+
+    // Open the per-repo actions menu, then choose Edit.
+    const actionsBtn = page
+      .getByRole('button', { name: new RegExp(`repository actions for`, 'i') })
+      .first();
+    await expect(actionsBtn).toBeVisible({ timeout: 10000 });
+    await actionsBtn.click();
+
+    const editItem = page.getByRole('menuitem', { name: /edit/i }).first();
+    await expect(editItem).toBeVisible({ timeout: 5000 });
+    await editItem.click();
+
+    const dialog = page.getByRole('dialog');
+    await expect(dialog).toBeVisible({ timeout: 5000 });
+    return dialog;
+  }
+
+  test('#412 + #413 toggling upstream-auth view to edit moves focus to the labelled auth-type select', async ({ page }) => {
+    const dialog = await openRemoteEditDialog(page);
+
+    // The remote repo has no auth configured -> the view shows a Configure button.
+    const toggleBtn = dialog.locator('#edit-upstream-auth-toggle');
+    await expect(toggleBtn).toBeVisible({ timeout: 5000 });
+    await toggleBtn.click();
+
+    // After toggling, focus must land on the first control of the edit view:
+    // the auth-type select trigger, which must carry a programmatic label.
+    const authTypeSelect = dialog.locator('#edit-upstream-auth-type');
+    await expect(authTypeSelect).toBeVisible({ timeout: 5000 });
+    await expect(authTypeSelect).toBeFocused({ timeout: 5000 });
+
+    // #413: the select has an associated label (it is a labelable control id).
+    const label = dialog.locator('label[for="edit-upstream-auth-type"]');
+    await expect(label).toBeVisible();
+
+    await dialog.getByRole('button', { name: /^cancel$/i }).first().click();
+  });
+
+  test('#410 saving upstream auth announces the outcome via an in-dialog live region', async ({ page }) => {
+    const dialog = await openRemoteEditDialog(page);
+
+    // The live region exists and is wired as an aria-live status region.
+    const liveRegion = dialog.getByTestId('upstream-auth-status');
+    await expect(liveRegion).toHaveAttribute('aria-live', /polite|assertive/);
+
+    // Configure -> choose bearer -> enter a token -> Save Authentication.
+    await dialog.locator('#edit-upstream-auth-toggle').click();
+
+    const authTypeSelect = dialog.locator('#edit-upstream-auth-type');
+    await expect(authTypeSelect).toBeVisible({ timeout: 5000 });
+    await authTypeSelect.click();
+    await page.getByRole('option', { name: /bearer/i }).click();
+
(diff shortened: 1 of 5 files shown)


brandonrc (commented):
Accessibility review (WCAG 2.2 AA). This is the dedicated a11y PR; I verified it against the four linked issues. It is solid work and the targeted fixes are correct. A few gaps remain.

**Verified closing #410/#411/#412/#413**

- #411 (create dialog duplicate-key error): correct. `#create-key` sets `aria-invalid` and `aria-describedby="create-key-error"`, and the message carries `role="alert"` with the matching id. SR announces it and ties it to the field (1.3.1, 3.3.1, 4.1.3). Good.
- #412 (focus on auth view<->edit toggle): the rAF + `getElementById(...).focus()` effect moves focus to the revealed control and the `editAuthModeInitialized` guard avoids stealing the dialog's initial focus. Reasonable given SelectTrigger does not forward a ref (2.4.3).
- #410 (live region for upstream-auth save): correct. `role=status`/`aria-live=polite` on success, `role=alert`/`aria-live=assertive` on error, and `onMutate` resets to idle so a repeat success re-announces (4.1.3). Good.
- #413 (aria-required, auth-type label, actions-menu name): all present. Repo row trigger now has `aria-label`, the auth-type select has a real `<label htmlFor>` (4.1.2). Good.

**Gaps to consider (non-blocking)**

1. `aria-required` was added but `aria-invalid`/error-association was NOT extended to the SSO OIDC/LDAP/SAML forms. Those forms still surface validation only via toast (transient, not programmatically tied to the field). #413 scoped this as a "broader gap" pass, so flagging that required-field validation errors in the SSO forms remain un-associated (3.3.1). Worth a follow-up issue.

2. #412: focus is moved on toggle, but when the auth-type Select is closed via Escape or the view collapses back to "view" mode, confirm focus lands on `#edit-upstream-auth-toggle` (the effect covers `editAuthMode === "view"` so this looks handled) and is not dropped if `editRepo.upstream_auth_configured` flips the rendered branch. Manual SR pass recommended.

3. The success live region uses `aria-live=polite` but the node is always present and only its text changes, which is the correct pattern. However the empty string is rendered as `""` inside the div on idle; that is fine. No change needed, just confirming it is not an empty-but-busy region.

No blockers. The three core fixes are correct and verifiable. Not approving (review-only).

brandonrc (commented):
Test/E2E quality review (v1.2.0 batch). Verdict: genuine a11y coverage, no theater. This is the model the other admin-gated specs should follow.

**#411 duplicate-key alert.** Strong. It doesn't just check the error text exists; it asserts the WCAG wiring that was the actual bug: `role="alert"`, `aria-invalid="true"`, and crucially that `aria-describedby` on the input equals the alert's `id` (`expect(describedBy).toBe(errorId)`). A regression that rendered the error text without wiring `aria-describedby` (the pre-fix state) would fail this. No skip path.

**#412 focus management.** This is the assertion most specs in this batch lack: `await expect(authTypeSelect).toBeFocused()` after the toggle. The bug was focus dumped on `<body>`; this directly distinguishes fixed-vs-broken. The `requestAnimationFrame`-based focus move in the component is matched by Playwright's auto-retrying `toBeFocused`, so no arbitrary wait needed. Good.

**#410 live region.** Honest. It asserts the region has `aria-live` polite/assertive, then drives the real save and asserts `not.toBeEmpty()` AND `toHaveText(/updated|fail|error/i)`. The text-regex accepts either success or error, which is the right tolerance for a backend that might reject the dummy token, while still proving the live region announces *something* (the bug was silence). It does not vacuously pass on an empty region.

No `test.skip` escape hatches anywhere in this file, no brittle ordering. The component diff backs every selector (`#create-key-error`, `id="edit-upstream-auth-type"`, `label[for=...]`, `data-testid="upstream-auth-status"`). Not requesting changes.

brandonrc:

PR #460, opened 2026-06-02
fix(repositories): Maven grouped browser pagination, file click-through, and full file listing

### LINKED ISSUE #443: Pagination not shown during Maven repository browse in Grouped mode
### Description

Inside browsing page of a Maven repository, the pagination is not displayed when in Grouped mode and we only see the first 20 artifacts (out of many more)

### Environment Information

- Browser: Chrome
- Browser Version: 148.0.7778.179 (Official Build) (64-bit)
- Operating System: Win11
- Screen Resolution: 2560x1440
- Connection type: direct

### Web UI Version

container image tag dev - currently 1.1.3

### Backend Version

container image tag dev - currently 1.2.0-rc.2

### Expected Behavior

Should be able to navigate to page 2, 3, etc...

### Observed Behavior

No link displayed to allow navigation to other pages of the list

### Steps to Reproduce

1. Browse to root URL of your Artifact Keeper instance
2. if you have a public Maven repository, just browse into it - if not, login and then browse to it
3. Select Grouped mode
4. If you have more than 20 artifacts stored, you will see the first 20, and no ways to list the other remaining artifacts
### Browser Console Logs

```shell
No errors listed in the browser console
```

### Screenshots

Image

### Additional Context

_No response_

### Pre-submission Checklist

### LINKED ISSUE #444: Cannot open artifact details when browsing the grouped mode
### Description

When browsing a Maven repository in Grouped mode (vs Flat mode), clicking on a artifact item will not open its details.

This do work when browsing in Flat mode.

### Environment Information

Browser: Chrome
Browser Version: 148.0.7778.179 (Official Build) (64-bit)
Operating System: Win11
Screen Resolution: 2560x1440
Connection type: direct

### Web UI Version

1.1.3 dev tag

### Backend Version

1.2.0-rc.2 dev tag

### Expected Behavior

Clicking on a Maven artifact should open a detail info box

### Observed Behavior

Nothing happens when clicking on a Maven artifact when the view is in Grouped mode

### Steps to Reproduce

Browse to a Maven repository
Set is to Grouped display (vs Flat)
Click on a artifact group
Click on a artifact appearing just below the group

### Browser Console Logs

```shell
No error message in the browser console
```

### Screenshots

_No response_

### Additional Context

_No response_

### Pre-submission Checklist

### LINKED ISSUE #445: Maven repository browser does not list all files associated with a specific GAV
### Description

It happens that when deploying a pom file to a Maven repository, the Maven project will also deploy with the pom file, ZIP files.
Those ZIP files are not listed in the Maven repository, in flat mode nor in Grouped mode

### Environment Information

N/A

### Web UI Version

1.1.3

### Backend Version

1.2.0-rc.2

### Expected Behavior

All stored files along side the pom file, probably the hashes as well, should be able to be downloaded from the repository browser; and we should be able to see who uploaded each of them, when, their file size, their last download date

### Observed Behavior

With a pom file , which has been generated together with some ZIP files, the Maven repository is only listing the pom file, and none of the ZIP files. Those ZIP files exist in the storage backend (filesystem)

### Steps to Reproduce

1. With a Maven project generating ZIP files, generate and deploy the project artifacts into Artifact Keeper inside a Maven repository
2. In the Maven repository, search for the artifact
3. List all files displayed and verify you can list the pom file and the ZIP files

### Browser Console Logs

```shell

```

### Screenshots

_No response_

### Additional Context

_No response_

### Pre-submission Checklist

brandonrc:
## Summary

Fixes three related defects in the Maven repository browser's GROUPED view.

- **#443** Grouped mode rendered no pagination, so only the first 20 GAV components were visible even when many more existed. The grouped branch used `MavenComponentList`, which had a static "showing N of M" footer but no page controls. Extracted the pagination markup out of `DataTable` into a shared `DataTablePagination` component and wired it into `MavenComponentList`, driven by the same `page`/`pageSize` state the flat list already uses.

- **#444** File rows inside an expanded GAV group were plain text spans that did nothing on click (flat mode opens details fine). They are now buttons that reconstruct the Maven path (`groupId/artifactId/version/filename`) and open the artifact detail dialog via a new `onFileSelect` callback.

- **#445** Non-jar files (e.g. `.zip` deployed alongside the POM, and checksum sidecars) could not be opened in grouped mode. The grouped backend response already lists every filename per GAV; surfacing each as a clickable row exposes them with full metadata (size, downloads, uploader, dates) through the detail dialog. Also fixed `artifactsApi.get` to encode path segments individually so the backend wildcard route `/:key/artifacts/*path` matches reconstructed paths instead of 404ing on percent-encoded slashes.

Closes #443
Closes #444
Closes #445

## Test Checklist
- [x] Manually tested locally

Added unit coverage for grouped pagination, clickable file rows, full file listing (incl. `.zip`), the new `mavenFilePath` helper, and segment-wise path encoding in `artifactsApi.get`. New Playwright spec `e2e/suites/interactions/repositories/maven-grouped-browser.spec.ts` seeds GAVs via the API and asserts grouped pagination renders and navigates, that clicking a grouped file opens the detail dialog, and that all GAV files (including `.zip` and checksums) are listed.

## UI Changes
- [ ] Dark mode verified

File rows are now native `<button>` elements with `aria-label`s, so they are keyboard-focusable and announced. Pagination reuses the existing accessible `DataTablePagination` controls. The grouped list inherits the existing dark-mode and responsive Tailwind classes unchanged.

### DIFF
--- a/e2e/suites/interactions/repositories/maven-grouped-browser.spec.ts
+++ b/e2e/suites/interactions/repositories/maven-grouped-browser.spec.ts
@@ -1,0 +1,185 @@
+import { test, expect, type APIRequestContext } from '@playwright/test';
+
+/**
+ * Regression tests for the Maven repository browser in GROUPED mode
+ * (issues #443, #444, #445).
+ *
+ * #443 — pagination must render in grouped mode so users can reach page 2+
+ *        instead of being stuck on the first 20 components.
+ * #444 — clicking a file inside a GAV group must open the artifact detail
+ *        dialog (this already worked in flat mode).
+ * #445 — every file deployed for a GAV must be listed, including non-jar
+ *        files such as .zip and the checksum sidecars.
+ *
+ * Strategy mirrors artifact-download.spec.ts: seed artifacts through the API,
+ * then drive the real UI.  The repository `e2e-maven-local` is created by the
+ * E2E seed step.
+ */
+test.describe('Maven Grouped Browser (#443, #444, #445)', () => {
+  const REPO_KEY = 'e2e-maven-local';
+  // Unique groupId per run so parallel/retried runs don't collide and so the
+  // assertions are deterministic regardless of what else lives in the repo.
+  const RUN = `g${Date.now()}`;
+  const GROUP_PATH = `com/example/${RUN}`;
+  const GROUP_ID = `com.example.${RUN}`;
+
+  /** PUT one file at a full Maven coordinate path. */
+  async function put(
+    request: APIRequestContext,
+    relPath: string,
+    body: string,
+  ): Promise<void> {
+    const resp = await request.put(
+      `/api/v1/repositories/${REPO_KEY}/artifacts/${relPath}`,
+      { data: body, headers: { 'Content-Type': 'application/octet-stream' } },
+    );
+    expect(resp.ok(), `PUT ${relPath} failed: ${resp.status()}`).toBeTruthy();
+  }
+
+  /**
+   * Deploy `count` distinct GAV components (one artifact + pom each) so the
+   * grouped listing has more than one page at the default page size of 20.
+   * Returns the list of artifactIds created.
+   */
+  async function seedManyComponents(
+    request: APIRequestContext,
+    count: number,
+  ): Promise<string[]> {
+    const ids: string[] = [];
+    for (let i = 0; i < count; i++) {
+      const artifactId = `lib-${i}`;
+      const version = '1.0.0';
+      const base = `${GROUP_PATH}/${artifactId}/${version}/${artifactId}-${version}`;
+      await put(request, `${base}.jar`, `jar-${i}`);
+      await put(request, `${base}.pom`, `<project>${i}</project>`);
+      ids.push(artifactId);
+    }
+    return ids;
+  }
+
+  /**
+   * Deploy a single GAV with a POM, a JAR, and a ZIP plus a checksum sidecar
+   * so #445 can assert that non-jar files are listed.
+   */
+  async function seedGavWithZip(request: APIRequestContext): Promise<{
+    artifactId: string;
+    files: string[];
+  }> {
+    const artifactId = 'with-zip';
+    const version = '2.0.0';
+    const base = `${GROUP_PATH}/${artifactId}/${version}/${artifactId}-${version}`;
+    const files = [
+      `${artifactId}-${version}.pom`,
+      `${artifactId}-${version}.jar`,
+      `${artifactId}-${version}.zip`,
+      `${artifactId}-${version}.jar.sha1`,
+    ];
+    await put(request, `${base}.pom`, '<project>zip</project>');
+    await put(request, `${base}.jar`, 'jar-bytes');
+    await put(request, `${base}.zip`, 'zip-bytes');
+    await put(request, `${base}.jar.sha1`, 'abc123');
+    return { artifactId, files };
+  }
+
+  async function gotoGrouped(
+    page: import('@playwright/test').Page,
+  ): Promise<void> {
+    // `?view=grouped` forces grouped mode regardless of stored preference.
+    await page.goto(`/repositories/${REPO_KEY}?view=grouped`);
+    await page.waitForLoadState('domcontentloaded');
+    const groupedBtn = page.getByTestId('toggle-grouped');
+    await expect(groupedBtn).toBeVisible({ timeout: 10000 });
+    if ((await groupedBtn.getAttribute('aria-pressed')) !== 'true') {
+      await groupedBtn.click();
+    }
+  }
+
+  test('#443: grouped mode renders pagination when there are many components', async ({
+    page,
+    request,
+  }) => {
+    // 25 components > default page size of 20, so a second page must exist.
+    await seedManyComponents(request, 25);
+
+    await gotoGrouped(page);
+
+    const list = page.getByTestId('maven-component-list');
+    await expect(list).toBeVisible({ timeout: 15000 });
+
+    // The pagination control must be present (the bug: it was never rendered).
+    const pagination = page.getByTestId('data-table-pagination');
+    await expect(pagination).toBeVisible();
+
+    // "Page 1 of N" with N >= 2 proves more than one page exists.
+    await expect(pagination).toContainText(/page 1 of [2-9]\d*/i);
+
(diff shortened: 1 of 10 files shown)


brandonrc (commented):
Accessibility review (WCAG 2.2 AA).

**Clickable file rows are real controls (the main question): yes.**
The new per-file rows in `maven-component-list.tsx` are real `<button type="button">` elements, not clickable divs, with `aria-label={`Open details for ${filename}`}` and `focus-visible` styling. They are keyboard operable and have an accessible name (4.1.2, 2.1.1). Good.

**Positive findings**
- The GAV disclosure trigger is a `<button aria-expanded>`; expand/collapse state is exposed (4.1.2).
- File and component lists keep `role="list"`; the FileIcon is `aria-hidden`. Correct.
- `DataTablePagination` reuses the shared control with named prev/next buttons.

**Gaps**

1. Moderate (2.4.7 / 3.2 focus visibility): the file-row button uses `focus-visible:bg-muted/50 focus-visible:outline-none`. Removing the outline and replacing it with only a faint background tint (`bg-muted/50`) is a weak focus indicator, especially on the muted text color, and may not meet WCAG 2.2 SC 2.4.11 Focus Appearance / 2.4.7 Focus Visible against the row background. Prefer keeping a visible ring (e.g. `focus-visible:ring-2 focus-visible:ring-ring`) rather than `outline-none` + background only.

2. Minor: the `<button>` accessible name is `Open details for {filename}`, but the visible text is just the filename. That is fine (name contains visible label, 2.5.3). No change needed.

3. Verify the new file detail dialog opened via `showDetailByPath` returns focus to the originating file button on close (2.4.3). The dialog is the shared artifact detail dialog, so this likely already works, but it is worth a manual SR/keyboard pass since the trigger is now a nested button inside a collapsible.

No blocker. Recommend restoring a real focus ring on the file-row button.

brandonrc (commented):
Test/E2E quality review (v1.2.0 batch). Verdict: real coverage at both layers; e2e has no lazy skip paths.

**#443 pagination unit.** Good and tight. `renders pagination controls when total is provided` asserts both `data-table-pagination` present AND `page 1 of 3` / `1-20 of 42`. The negative case (`total undefined` => no pagination) is covered. `onPageChange called with 2` proves the wiring. The prior tests asserted a "showing N of M" footer; the bug was that grouped mode never rendered real pagination, and these tests now assert the actual control, so they distinguish fixed-vs-broken.

**#444/#445 unit.** `onFileSelect` is asserted to receive the reconstructed Maven path `org/junit/.../...jar` plus filename, proving the click-to-detail wiring. The non-jar listing test explicitly checks `.zip` and `.jar.sha1` rows render, which is exactly the #445 bug.

**e2e.** Notably this is the only repo-browser spec in the batch with no admin-gating skip. It seeds 25 components and asserts `page 1 of [2-9]\d*` then navigates to `page 2 of` (real pagination assertion, not just presence). The #444 test asserts clicking the .jar opens a dialog with a Download URL field. The #445 test asserts all 4 files visible AND the `4 files` count badge. Selectors (`data-testid="maven-component-row"`, `data-gav`, `toggle-grouped` with `aria-pressed`, `data-table-pagination`) are all added in the component diff, so these are not testing against non-existent attributes.

Minor: `page.getByText(/page 1 of [2-9]\d*/i)` assumes seeded set stays >20 components; since groupId is per-run-unique this is deterministic. One `waitForTimeout(500)` after expand is an arbitrary wait but low risk. Not requesting changes.

brandonrc (commented):
Security/API review (path traversal + encoding lens).

Net: safe and a genuine correctness improvement. No new traversal exposure.

Per-segment encoding in artifactsApi.get — CORRECT. Switching from `encodeURIComponent(artifactPath)` (which would `%2F`-encode separators and break the Axum `*path` wildcard) to `artifactPath.split('/').map(encodeURIComponent).join('/')` preserves slashes while escaping reserved characters within each segment (space -> %20, etc.). This does not widen traversal exposure: `..` as a literal segment stays `..` (encodeURIComponent does not touch `.`), so the path sent is no more dangerous than before — and traversal prevention for the wildcard route is the backend's responsibility, unchanged here. Spaces/`#`/`?`/`%` in segments are now correctly escaped, which actually reduces request-smuggling/ambiguous-parse risk versus the old whole-string encode.

One thing to verify outside this PR (not blocking): confirm the backend `/:key/artifacts/*path` handler canonicalizes/rejects `..` and absolute paths before touching storage. The client now faithfully forwards multi-segment paths, so the backend remains the sole traversal guard. If you want, I can spot-check the Rust handler.

No request-changes.

brandonrc (commented):
Frontend/React correctness review. No blockers. Clean extraction and a correct encoding fix.

### Correctness notes

- **`artifactsApi.get` per-segment encoding** is the right fix: splitting on `/` and `encodeURIComponent`-ing each segment preserves the Axum wildcard (`/:key/artifacts/*path`) while still escaping reserved chars within a segment. The unit test asserts spaces become `%20` and no `%2F` appears. Good. Since `get` runs against `getActiveInstanceBaseUrl()`, the grouped-mode detail lookup also works under a proxied remote instance (a nice contrast with #457's absolute-URL gap).

- **`DataTablePagination` extraction** preserves behavior: the render-nothing-when-no-handlers guard moved inside the component (`if (!onPageChange && !onPageSizeChange) return null`), and `DataTable` now always renders it but the guard keeps the old semantics. No regression.

- **Grouped pagination visibility**: `MavenComponentList` renders pagination whenever `typeof total === "number"`, and `repo-detail-content` always passes `onPageChange`/`onPageSizeChange`, so the grouped view now matches the flat list (controls always shown, even on a single page). Consistent with `DataTable`. Fine.

- **`showDetailByPath`** is correctly memoized on `[repoKey]` and has user-facing error handling (toast on fetch failure). Clickable file rows are real `<button>`s with `aria-label`s. The reconstructed path goes through the fixed `get`, closing the #444/#445 loop.

### Nit
- `maven-component-path.ts` `mavenFilePath` and `src/lib/maven.ts` `mavenFilePath` (from #458) are near-duplicate helpers in two locations. Not a defect, but if both PRs land, consider consolidating on the shared `src/lib/maven.ts` version to avoid drift. Merge-sequencing overlap with #458 on `maven-component-list.tsx` is known.

Well-tested (unit + e2e). LGTM.

brandonrc:

PR #461, opened 2026-06-02
test(e2e): verify remote-repo cached artifacts show in the browser

### LINKED ISSUE #424: Artifacts no longer show in UI
### Description

When using remote repositorities, artifacts are no longer showing in the repo screen in the UI. The packages are brought in as I can see my S3 storage filling up. This means we can't run scans on them etc.

### Environment Information

- Browser: (e.g., Chrome, Firefox, Safari, Edge)
- Browser Version:
- Operating System:
- Screen Resolution:
- Connection type: (direct, behind proxy)

### Web UI Version

1.1.3

### Backend Version

1.2.0-rc.1

### Expected Behavior

When pulling in artifacts from any source, I would expect the artifacts to show in UI so we can trigger scans  and see what is in repo.

### Observed Behavior

When using remote repositorities, artifacts are no longer showing in the repo screen in the UI. The packages are brought in as I can see my S3 storage filling up. This means we can't run scans on them etc.

### Steps to Reproduce

1) Run pnpm install
2) After packages install, check repo in  UI, shows 0 artifacts/packages
3) Check S3 storage, size is not empty.

### Browser Console Logs

```shell

```

### Screenshots

_No response_

### Additional Context

_No response_

### Pre-submission Checklist

brandonrc:
## Summary

Fixes the v1.2.0 regression reported in #424: when using remote (proxy) repositories, artifacts pulled through stopped appearing in the repo browser. Storage filled up (bytes were written) but the UI listed 0 artifacts, so users could not browse or scan them.

Root cause is in the backend. #1278 / #1280 stopped recording proxy-cached items in the `artifacts` table to fix a filesystem doubled-prefix storage path bug. The acknowledged tradeoff was that remote-cached items no longer surfaced in `GET /api/v1/repositories/{key}/artifacts`, which is exactly what the repo browser lists from. The backend fix (artifact-keeper#1567, tracking artifact-keeper#1548) reconstructs the listing for remote repos from the proxy cache, preserving the #1278 fix.

The web client already consumes the corrected data through the same endpoint, so no web code change is required. This PR adds the Playwright regression test for the user-visible flow:

- Pulls a small package (`is-odd`) through the seeded `e2e-npm-remote` proxy.
- Asserts the cached entry appears in the listing API with real size and checksum.
- Asserts the entry shows as a row in the Artifacts tab of the repo detail page.

The spec skips gracefully if the upstream registry is unavailable (upstream availability is not what's under test). It runs in CI against the 1.2.0 backend image.

## Test Checklist
- [x] Manually tested locally — eslint and tsc pass on the new spec

## UI Changes
- [x] N/A - no UI changes

This is a test-only change; the behavior fix lands in the backend (artifact-keeper#1567).

Closes #424

### DIFF
--- a/e2e/suites/interactions/repositories/remote-cached-artifacts.spec.ts
+++ b/e2e/suites/interactions/repositories/remote-cached-artifacts.spec.ts
@@ -1,0 +1,115 @@
+import { test, expect } from '@playwright/test';
+
+/**
+ * Regression test for issue #424 — "Artifacts no longer show in UI".
+ *
+ * When a remote (proxy) repository pulls a package through to its upstream,
+ * the cached object must appear in the repository artifact listing so it can
+ * be browsed and scanned. The v1.2.0 regression (backend #1278 / #1280)
+ * stopped recording proxy-cached items in the `artifacts` table to fix a
+ * filesystem storage-path bug; the side effect was that remote-cached items
+ * vanished from:
+ *
+ *   GET /api/v1/repositories/{key}/artifacts
+ *
+ * and therefore from the repo browser UI, even though the bytes were written
+ * to storage. The backend now reconstructs the listing for remote repos from
+ * the proxy cache (backend #1548), so this verifies the user-visible flow:
+ * pull a package through the seeded NPM remote, then confirm it shows up in
+ * both the listing API and the Artifacts tab of the repo detail page.
+ *
+ * Uses the seeded `e2e-npm-remote` repository (proxies registry.npmjs.org,
+ * see e2e/setup/seed-data.ts). The NPM proxy routes are mounted at
+ * `/npm/{repo_key}/...`; fetching package metadata caches an entry under the
+ * package path, which the listing surfaces with `path === <package>`.
+ */
+test.describe('Remote repository cached artifacts (#424)', () => {
+  const REPO_KEY = 'e2e-npm-remote';
+  // A tiny, stable package with no dependencies — cheap to pull through.
+  const PACKAGE = 'is-odd';
+
+  /**
+   * Pull a package through the NPM proxy so the backend caches it. Returns
+   * true once the proxy responds successfully; remote upstream hiccups should
+   * skip the test rather than fail it (the listing behavior, not upstream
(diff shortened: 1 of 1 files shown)


brandonrc (commented):
Test/E2E quality review (v1.2.0 batch). Verdict: the assertions ARE good when they run, but both tests sit behind an upstream-availability skip that will frequently swallow the whole fix in CI. This is the one I am most skeptical of.

**The skip problem.** Both tests call `pullThroughProxy()` and, if `!resp.ok()`, do `test.skip(true, 'Upstream NPM registry unavailable...')`. The bug under test (#424: remote-cached items vanished from the listing) can ONLY manifest after a successful pull-through. So whenever registry.npmjs.org is unreachable, slow, or rate-limited from the CI runner, both tests go green having asserted nothing about the fix. For a release-gate e2e that depends on a third-party registry, this is a real "perpetually green without testing" risk, not a theoretical one. The comment frames upstream availability as out of scope, but the practical effect is that the fix's only e2e coverage is conditional on a flaky external dependency.

**When it does run, the assertions are honest.** The listing-API test asserts `items.length > 0` (the regression was an empty listing), finds the package by path/name, and asserts `size_bytes > 0` and `checksum_sha256` truthy, proving the reconstructed-from-sidecar metadata is real, not stubbed. The UI test asserts a row containing the package is visible in the Artifacts tab. These genuinely distinguish fixed-vs-broken.

**Recommendation.** Reduce the external dependency: either (a) seed the proxy cache deterministically (PUT a sidecar+content pair into `proxy-cache/<key>/...` via storage, or a fixture upstream served by the test harness) so the listing assertion always runs, or (b) point the remote at a local stub registry. Pair this with the backend gap I flagged on #1567 (no unit test exercises `list_cached_artifacts` against a storage backend) — right now, between the two PRs, there is a plausible CI run where the entire #424 fix is asserted by nothing. Strongly suggest fixing the determinism before relying on this as release-gate coverage. Not formally requesting changes, but this is the weakest spec in the batch.

brandonrc:

PR #463, opened 2026-06-02
feat(search): surface OpenSearch capabilities in the search UI

### LINKED ISSUE #269: Explore: search UI changes needed for OpenSearch backend
Backend: artifact-keeper/artifact-keeper#462

The backend may add OpenSearch as an alternative to Meilisearch for search indexing. Explore whether the frontend search UI needs any changes:

1. Does the search API response format change between Meilisearch and OpenSearch?
2. Are there new search capabilities (faceted search, filters, aggregations) that the UI should expose?
3. Does the system config endpoint expose which search backend is active?
4. Should the admin settings show search backend health/status?

This is an exploration ticket. No implementation needed until the backend work is scoped.

brandonrc:
Closes #269

## Summary

The backend migrated search indexing from Meilisearch to OpenSearch in 1.2.0 (artifact-keeper/artifact-keeper#462). This PR is the investigation plus the concrete, high-value UI changes that migration calls for, answering the four questions in #269.

### Investigation findings

1. Response format: the `/search/advanced` shape is unchanged (items, pagination, facets) and the SDK types match. No breaking client changes were needed, but the UI was discarding parts of the existing response.
2. New capabilities now worth exposing:
   - Server-side sort honors `sort_by` plus `sort_order` (asc/desc) per field, and `sort_by` accepts `downloads` (alias `download_count`). The UI only ever sent `sort_by` and never the order or downloads.
   - `/search/advanced` returns facet aggregations (formats, repositories, content_types). The API adapter was dropping them entirely.
   - Results carry highlight snippets; the type modeled them but nothing rendered them.
3. Health endpoint: `/health` now exposes the search backend under `checks.opensearch` (it was `checks.meilisearch`). The web health adapter still read `meilisearch`, so after the backend upgrade the dashboard "Search Engine" card silently vanished.
4. Admin/dashboard status: the dashboard already renders a search-engine health card; the fix above restores it.

### Implemented changes

Search results UI (`search-content.tsx`):
- Relevance is now the default sort and sends no `sort_by`, letting OpenSearch apply its own ranking. Added a Downloads sort option.
- Added a sort-direction toggle that sends `sort_order=asc|desc`. It is disabled while sorting by relevance (no direction applies).
- Removed the client-side re-sort that only reordered the current page and contradicted the server ordering across pages.
- Render highlight snippets, parsing the `<em>` markers into React nodes instead of injecting raw HTML.
- Surface the formats and repositories facets as clickable refine chips that re-issue the search with the facet applied, plus a clear-filters control.

API + types:
- `searchApi.advancedSearch` now returns facets (`AdvancedSearchResult` widens `PaginatedResponse`).
- Health adapter reads `checks.opensearch` and falls back to `checks.meilisearch`; the dashboard reads whichever is present.

### Deferred (needs product input, noted not implemented)

- Saved searches and the broader property/GAVC filter model already exist as types but are out of scope here.
- Exposing the active search backend name (Meilisearch vs OpenSearch) as explicit UI copy was left out; the health card label stays the neutral "Search Engine". Surfacing engine identity or a reindex trigger in admin settings can follow once product decides whether to show it.

## Test Checklist
- [ ] Manually tested locally

## UI Changes
- [ ] Dark mode verified

The new sort controls and facet chips use semantic roles (`combobox` with `aria-label`, `aria-pressed` on facet toggles, `aria-label` describing the sort direction). New e2e spec: `e2e/suites/interactions/search/opensearch-search.spec.ts`.

### DIFF
--- a/e2e/suites/interactions/search/opensearch-search.spec.ts
+++ b/e2e/suites/interactions/search/opensearch-search.spec.ts
@@ -1,0 +1,198 @@
+import { test, expect } from '@playwright/test';
+
+/**
+ * Search UI coverage for the OpenSearch backend (issue #269, v1.2.0).
+ *
+ * The backend migrated search indexing to OpenSearch in 1.2.0. This suite
+ * exercises the capabilities the UI now exposes on top of that:
+ *
+ * 1. Relevance is the default sort (OpenSearch ranks results, no explicit
+ *    sort_by is sent). Date, Name, Size, and Downloads are explicit sorts.
+ * 2. A sort-direction toggle sends sort_order=asc|desc, and is disabled while
+ *    sorting by relevance (relevance has no direction).
+ * 3. The advanced search request carries sort_by and sort_order when an
+ *    explicit field sort is chosen.
+ * 4. Facets returned by the backend render as clickable refine chips that
+ *    re-issue the search with the facet applied.
+ *
+ * Runs in CI against the :1.2.0 backend image. Tests that need indexed data
+ * seed it through the public artifact API, mirroring artifact-download.spec.ts.
+ */
+test.describe('OpenSearch Search UI', () => {
+  const REPO_KEY = 'e2e-maven-local';
+
+  async function uploadArtifact(
+    request: import('@playwright/test').APIRequestContext,
+    name: string
+  ): Promise<string> {
+    const path = `e2e/opensearch/${name}-${Date.now()}.txt`;
+    const resp = await request.put(
+      `/api/v1/repositories/${REPO_KEY}/artifacts/${path}`,
+      {
+        data: `OpenSearch indexing fixture for ${name}`,
+        headers: { 'Content-Type': 'application/octet-stream' },
+      }
+    );
+    expect(resp.ok(), `Upload failed: ${resp.status()}`).toBeTruthy();
+    const body = await resp.json();
+    return body.path || path;
+  }
+
+  test('sort menu offers relevance, date, name, size, and downloads', async ({
+    page,
+  }) => {
+    await page.goto('/search');
+    await page.getByPlaceholder('e.g., react, lodash').fill('test');
+    await page.getByRole('button', { name: /^search$/i }).first().click();
+
+    // Results header (results or empty state) must be present before the
+    // sort control renders.
+    await expect(page.getByText(/results/i).first()).toBeVisible({
+      timeout: 15000,
+    });
+
+    const sortSelect = page.getByRole('combobox', { name: /sort by/i });
+    await expect(sortSelect).toBeVisible({ timeout: 10000 });
+    await sortSelect.click();
+
+    // Radix Select renders options into a listbox popover.
+    await expect(
+      page.getByRole('option', { name: /relevance/i })
+    ).toBeVisible({ timeout: 10000 });
+    await expect(page.getByRole('option', { name: /^date$/i })).toBeVisible();
+    await expect(page.getByRole('option', { name: /^name$/i })).toBeVisible();
+    await expect(page.getByRole('option', { name: /^size$/i })).toBeVisible();
+    await expect(
+      page.getByRole('option', { name: /downloads/i })
+    ).toBeVisible();
+  });
+
+  test('sort direction toggle is disabled for relevance and enabled for field sorts', async ({
+    page,
+  }) => {
+    await page.goto('/search');
+    await page.getByPlaceholder('e.g., react, lodash').fill('test');
+    await page.getByRole('button', { name: /^search$/i }).first().click();
+    await expect(page.getByText(/results/i).first()).toBeVisible({
+      timeout: 15000,
+    });
+
+    // Default sort is relevance, so the direction toggle is disabled (there is
+    // no ascending/descending relevance).
+    const toggle = page.getByRole('button', { name: /sort (descending|ascending)/i });
+    await expect(toggle).toBeVisible({ timeout: 10000 });
+    await expect(toggle).toBeDisabled();
+
+    // Switch to an explicit field sort; the direction toggle becomes enabled.
+    const sortSelect = page.getByRole('combobox', { name: /sort by/i });
+    await sortSelect.click();
+    await page.getByRole('option', { name: /^date$/i }).click();
+    await expect(toggle).toBeEnabled({ timeout: 10000 });
+  });
+
+  test('advanced search request carries sort_by and sort_order for field sorts', async ({
+    page,
+  }) => {
+    await page.goto('/search');
+    await page.getByPlaceholder('e.g., react, lodash').fill('test');
+    await page.getByRole('button', { name: /^search$/i }).first().click();
+    await expect(page.getByText(/results/i).first()).toBeVisible({
+      timeout: 15000,
+    });
+
+    // Choose Size sort, which sends sort_by=size_bytes. Capture the next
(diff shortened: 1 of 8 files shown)


brandonrc (commented):
Accessibility review (WCAG 2.2 AA) of the OpenSearch search UI (sort toggle, facet chips, highlight rendering).

**Positive findings**
- Sort select: `SelectTrigger` now carries `aria-label="Sort by"` (4.1.2). Good.
- Sort-direction toggle: real `<Button>` with a descriptive, state-aware `aria-label` ("Sort descending, switch to ascending" / vice-versa) and `disabled` when sort is relevance. The icon is decorative-only and the name carries the meaning (4.1.2, 1.4.1). This is a good pattern, not icon-only.
- Facet chips are real `<button type="button">` with `aria-pressed={...}` reflecting selection state, and the visible Badge text (value + count) gives the button an accessible name (4.1.2, 2.1.1). The Clear filters button has visible text. Good.
- Highlight rendering: `renderHighlight` parses `<em>` snippets into React `<mark>` nodes instead of `dangerouslySetInnerHTML`. Both an a11y win (real text nodes, SR-readable) and an XSS win.

**Gaps**

1. Moderate (4.1.3 Status Messages): when a sort change, facet toggle, or clear re-runs the search and the result count changes, nothing is announced to screen readers. The results region updates silently. Add an `aria-live="polite"` status node (e.g. "42 results" / "Showing results for ...") near the results header so SR users learn the list changed after operating the sort/facet controls. Right now the visible "results" text exists but is not in a live region.

2. Moderate (1.4.1 Use of Color): the active facet is conveyed by `Badge variant="default"` vs `"secondary"` (a color change). `aria-pressed` covers screen readers, but sighted low-vision users may not distinguish default vs secondary badge colors. Consider an additional non-color affordance (a check icon or a visible "x to remove" on the active chip). The `aria-pressed` makes this non-blocking for SR.

3. Minor: `<mark>` is styled `bg-transparent ... text-foreground`, i.e. the highlight is conveyed only by `font-semibold`. Bold-only emphasis is acceptable, and `<mark>` carries semantic meaning, so this is fine. No change required.

4. Minor: facet chip button has no `type`-related issue, but confirm the chip's accessible name reads sensibly ("maven 12") to SR users; consider `aria-label={`${f.value}, ${f.count} results`}` for clarity.

No blocker. Item 1 (live region on result updates) is the main one to address. Review-only, not approving.

brandonrc (commented):
Test/E2E quality review (v1.2.0 batch). Verdict: the request-interception assertions make this genuinely strong; one soft facet test.

**sort_by/sort_order request assertions (#269).** This is the right way to test a search-param fix and it's the best in the batch. `advanced search request carries sort_by and sort_order` uses `page.waitForRequest` to capture the actual outgoing `/api/v1/search/advanced` call and asserts `sort_by=size_bytes` + `sort_order=desc`, then flips the toggle and captures a second request asserting `sort_order=asc`. The companion `relevance sort omits sort_by` asserts `params.get('sort_by')` is null. Together these prove the OpenSearch sort contract in both directions, and they would fail against any implementation that hardcoded a sort or omitted the param. No skip path on these.

**direction toggle disabled-for-relevance.** Asserts `toggle.toBeDisabled()` under relevance and `toBeEnabled()` after picking Date. Real behavioral assertion, not just presence.

**sort menu options.** Presence-only (relevance/date/name/size/downloads visible), which is fine as a smoke check.

**facets test.** This is the weak one. It is wrapped in `if (facets.isVisible().catch(()=>false))` with an else branch that only pushes a `test.info().annotation` "not yet indexed". So if OpenSearch indexing lags (which it will in CI, indexing is async), the entire facet refine/clear-filters assertion never runs and the test is green having tested nothing about facets. Given indexing latency this branch will frequently be the one taken. The note acknowledges it but a perpetually-skipped assertion is theater. Suggest either: poll/retry the search until facets appear (bounded), or assert facets against a pre-seeded indexed fixture so the refine flow actually executes in CI. Flag, but the sort coverage carries the PR. Not requesting changes.

brandonrc (commented):
Security review (XSS lens on search-result highlight rendering).

renderHighlight — SAFE, no XSS. No `dangerouslySetInnerHTML` and no `innerHTML` anywhere in the change (grepped). The snippet is split on `/(<em>.*?<\/em>)/g`, and each part is either matched as `<em>...</em>` (rendering only the captured group `match[1]` inside a React `<mark>`) or rendered as a plain `<span>{part}</span>`. In both branches the content goes through React's text-node path, which HTML-escapes it. So a malicious indexed value such as `` arriving from OpenSearch is rendered as inert text, not parsed as markup. The only markup that becomes real DOM is the static `<mark>`/`<span>` elements this code creates.

Edge cases reviewed: a literal `<em>` the user managed to index would be stripped of its tags and shown as emphasized text (cosmetic only, not an injection). Unbalanced or nested `<em>` markers fall into the plain-text branch and are escaped. No event handlers or URLs are derived from the snippet.

No request-changes. This is the correct pattern (parse-then-render-as-nodes) and a clear improvement over raw HTML insertion.

brandonrc (commented):
Frontend/React correctness review. No blockers. Strong PR: removes a genuinely buggy client-side re-sort, renders highlights safely, and adapts facets/health cleanly.

### Correctness notes

- **Removing the client-side re-sort is the right call.** The old `useMemo` only reordered the current page and contradicted the server ordering across pages. Rendering `searchResults.items` directly is correct now that OpenSearch owns ranking and `sort_by`/`sort_order` are sent.

- **`renderHighlight` is XSS-safe.** Splitting on `(<em>.*?</em>)` and rendering matched groups as `<mark>` text nodes (no `dangerouslySetInnerHTML`) means any other markup or injected content in the snippet renders as inert text. Good.

- **`relevance` => omit `sort_by`** is handled consistently: `sort_by` undefined and `sort_order` only sent when a real sort field is active. The direction toggle is correctly disabled for relevance. Query key includes `sortField`, `sortOrder`, `facetFormat`, `facetRepository`, so the cache invalidates on every dimension that changes the request.

- **Health adapter** (`opensearch ?? meilisearch`, mapped onto both keys) plus the dashboard's `opensearch ?? meilisearch` read correctly restores the "Search Engine" card across backend versions. `search.ts` facet adaptation defensively treats `sdk.facets` as possibly-undefined even though the SDK types it required, which is fine.

- `advancedSearch` still goes through the generated SDK; I confirmed the SDK query type accepts `sort_order` and the SDK models `FacetsResponse`/`FacetValue`, so the new params/fields are not silently dropped.

### Blocker-adjacent merge conflict with #458 (coordinate the rebase)

This PR's `gavc` branch in `buildSearchParams` still sends `path`/`name`/`version`:
```tsx
case "gavc":
  return { ...common, path: gavcValues.groupId, name: gavcValues.artifactId, version: gavcValues.version };
```
That is exactly the broken behavior #458 fixes (the advanced-search endpoint ignores those params; GAV must be folded into a single full-text `query` scoped to `format: "maven"`). If this PR merges after #458 it will silently regress GAV search. Whichever lands second must keep #458's `buildMavenSearchQuery` mapping for the gavc tab. Please rebase carefully and re-run the #458 e2e (`maven-gav-search-pom.spec.ts`) after merge.

### Nit
- `facetRepository ?? (packageValues.repository || undefined)` etc. give a selected facet precedence over the form field. Intended per the comment; just confirm a user can still clear a facet to fall back to the form value (the "Clear filters" control handles this).

LGTM aside from the #458 rebase coordination.

brandonrc:

PR #464, opened 2026-06-02
Add package age policy and upload size limit configuration UIs

### LINKED ISSUE #189: Show and configure repository upload size limit in admin settings
### Problem

There is no way to view or configure the maximum artifact upload size from the web UI. The backend will support a `MAX_UPLOAD_SIZE` environment variable (artifact-keeper/artifact-keeper#487), but admins should also be able to see the current limit and potentially adjust it from the settings page.

### Proposed Solution

1. Display the current max upload size on the Settings or System Health page
2. Optionally allow admins to view the configured limit (read-only from the backend's `/api/v1/system/config` or similar endpoint)
3. Show a helpful message on upload failure when the body limit is exceeded (instead of a generic 413 error)

### References

- Backend issue: artifact-keeper/artifact-keeper#487
- PR #484 on backend (closed): proposed removing the limit entirely

### LINKED ISSUE #265: Add package age policy configuration UI
Backend: artifact-keeper/artifact-keeper#709

The backend supports a policy to disallow packages released less than X time ago. The frontend needs a way to configure this policy in the admin or repo settings.

brandonrc:
## Summary

Adds two repository configuration surfaces for v1.2.0.

**Package age policy (#265)** — a "Package Age Policy" section on the repo Settings tab. When enabled, freshly published packages pulled through a remote repository are held in quarantine for a configurable cooldown window (hours or days). This gives time to flag a compromised upstream release before it reaches clients, mitigating supply-chain attacks. The toggle and duration are written to `PATCH /api/v1/repositories/{key}` as `quarantine_enabled` and `quarantine_duration_minutes` (the per-repo keys the backend stores in `repository_config`, per artifact-keeper/artifact-keeper#709). These fields are not in the generated SDK yet, so the call goes through `apiFetch`, matching the existing pattern for upstream-auth and test-upstream.

**Upload size limit (#189)** — the admin Settings -> Storage tab previously showed Max Upload Size read-only. It is now editable (numeric value + MB/GB unit selector + Save), persisted via `POST /api/v1/admin/settings`. Because that endpoint replaces the whole `SystemSettings` object, the save reads current settings first and sends them back with only `max_upload_size_bytes` changed, so sibling values are preserved. The repo Settings tab also surfaces the effective limit read-only so repo owners can see the instance-wide ceiling that applies to uploads.

## Test Checklist
- [x] Manually tested locally

## UI Changes
- [x] Dark mode verified
- [ ] N/A - no UI changes

Closes #265
Closes #189

### DIFF
--- a/e2e/suites/interactions/admin/upload-size-limit.spec.ts
+++ b/e2e/suites/interactions/admin/upload-size-limit.spec.ts
@@ -1,0 +1,117 @@
+import { test, expect } from '@playwright/test';
+
+/**
+ * Configuration flow for the repository upload size limit (issue #189).
+ *
+ * The admin Settings -> Storage tab surfaces the current max upload size and
+ * lets an admin change it. The control reads from and writes to:
+ *   GET  /api/v1/admin/settings
+ *   POST /api/v1/admin/settings   (full SystemSettings body)
+ *
+ * Runs in CI against backend :1.2.0.
+ */
+test.describe('Admin - Upload Size Limit', () => {
+  test('admin settings expose and accept a new max upload size', async ({ request }) => {
+    const getResp = await request.get('/api/v1/admin/settings');
+    expect(getResp.ok(), `GET settings failed: ${getResp.status()}`).toBeTruthy();
+
+    const settings = await getResp.json();
+    expect(settings).toHaveProperty('max_upload_size_bytes');
+    const original = settings.max_upload_size_bytes;
+
+    // Send the whole object back with only the upload size changed, matching
+    // how the UI persists it (the POST replaces the full SystemSettings).
+    const target = 2 * 1024 * 1024 * 1024; // 2 GiB
+    const putResp = await request.post('/api/v1/admin/settings', {
+      data: { ...settings, max_upload_size_bytes: target },
+      headers: { 'Content-Type': 'application/json' },
+    });
+    expect(putResp.ok(), `POST settings failed: ${putResp.status()}`).toBeTruthy();
+
+    const verify = await request.get('/api/v1/admin/settings');
+    const after = await verify.json();
+    expect(after.max_upload_size_bytes).toBe(target);
+
+    // Restore the original value so the suite is idempotent.
+    await request
+      .post('/api/v1/admin/settings', {
+        data: { ...settings, max_upload_size_bytes: original },
+        headers: { 'Content-Type': 'application/json' },
+      })
+      .catch(() => {});
+  });
+
+  test('Storage tab shows an editable Max Upload Size control', async ({ page }) => {
+    await page.goto('/settings');
+    await page.waitForLoadState('domcontentloaded');
+    await page.waitForTimeout(2000);
+
+    const storageTab = page.getByRole('tab', { name: /storage/i }).first();
+    if (await storageTab.isVisible({ timeout: 5000 }).catch(() => false)) {
+      await storageTab.click({ force: true });
+      await page.waitForTimeout(1000);
+    }
+
+    // The Max Upload Size label and its editable number input must be present.
+    await expect(page.getByText('Max Upload Size').first()).toBeVisible({
+      timeout: 10000,
+    });
+
+    const sizeInput = page.getByLabel('Max Upload Size');
+    await expect(sizeInput).toBeVisible({ timeout: 5000 });
+    await expect(sizeInput).toBeEditable();
+
+    // The unit selector and a Save button accompany the input.
+    await expect(
+      page.getByLabel('Upload size unit')
+    ).toBeVisible({ timeout: 5000 });
+
+    const body = await page.textContent('body');
+    expect(body).not.toContain('Application error');
+  });
+
+  test('saving a changed upload size issues a settings POST', async ({ page }) => {
+    await page.goto('/settings');
+    await page.waitForLoadState('domcontentloaded');
+    await page.waitForTimeout(2000);
+
+    const storageTab = page.getByRole('tab', { name: /storage/i }).first();
+    if (await storageTab.isVisible({ timeout: 5000 }).catch(() => false)) {
+      await storageTab.click({ force: true });
+      await page.waitForTimeout(1000);
+    }
+
+    const sizeInput = page.getByLabel('Max Upload Size');
+    if (!(await sizeInput.isVisible({ timeout: 5000 }).catch(() => false))) {
+      test.skip(true, 'Upload size control not visible');
+      return;
+    }
+
+    // Capture the current value to restore later.
+    const originalValue = await sizeInput.inputValue();
+
+    await sizeInput.fill('512');
+
+    const postResponse = page.waitForResponse(
+      (r) =>
+        r.url().includes('/api/v1/admin/settings') &&
+        r.request().method() === 'POST',
+      { timeout: 10000 }
+    ).catch(() => null);
+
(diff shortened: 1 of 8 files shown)


brandonrc (commented):
Accessibility review (WCAG 2.2 AA) of the age-policy and upload-size forms. NOTE: this contains an a11y BLOCKER (would be a request-changes; posting as a comment because GitHub does not allow request-changes on one's own PR). Please treat item under "Blocker" as merge-gating.

**Positive findings**
- Age policy: `<section aria-labelledby="settings-age-heading">` tied to a real `<h3 id>`. The enable `Switch` has `<Label htmlFor="settings-age-enabled">`; the cooldown `<Input id="settings-age-duration">` has `<Label htmlFor>`. Toggle-gated disabling (`disabled={!ageEnabled}`) is correct and conveyed natively (1.3.1, 4.1.2, 3.3.2).
- Upload size (global Settings): `<Label htmlFor="max-upload-size">`, the unit `SelectTrigger` has `aria-label="Upload size unit"`, Save disabled until dirty. The read-only mirror on the repo settings tab uses `disabled` + `aria-label="Upload size limit"`. Good.

**Blocker (Serious, WCAG 3.3.1 Error Identification / 4.1.3 Status Messages)**
The cooldown validation error in `repo-settings-tab.tsx` is not programmatically associated with the input. The input sets `aria-invalid={ageInvalid}`, but the error paragraph:

    {ageInvalid ? (
      <p className="text-sm text-destructive">Enter a cooldown period of at least one ...</p>
    ) : ( ... )}

has no `id`, no `role="alert"`, and `#settings-age-duration` has no `aria-describedby` pointing at it. A screen-reader user who clears the field hears only "invalid" with no reason, and the new error text is never announced. Fix:

- give the error `<p>` a stable `id` (e.g. `settings-age-duration-error`) and `role="alert"`;
- set `aria-describedby={ageInvalid ? "settings-age-duration-error" : undefined}` on the cooldown `<Input>`.

Reuse the exact pattern PR #459 established for the create-dialog key error (`create-key-error` + `role=alert` + `aria-describedby`).

**Non-blocking**
- Minor: the helper/error `<p>` swaps between description and error text in the same slot. With `role="alert"` only on the error variant, this is fine; ensure the description variant does NOT carry `role="alert"` (it currently does not).
- Minor (4.1.3): upload-size Save success/failure is toast-only. Acceptable for an admin action.

Once the cooldown error is associated via `aria-describedby` + `role="alert"`, this is good to go. Not approving/merging.

brandonrc (commented):
Test/E2E quality review (v1.2.0 batch). Verdict: the unit helpers are well-covered; the e2e UI tests are soft (best-effort POST checks that pass even when nothing is saved).

**Unit tests (#189) — solid.** `bytesToUploadSize`/`uploadSizeToBytes` cover whole-GB, MB fallback, 0-as-no-limit (empty value), reverse conversion, invalid input (`""`, `-1`, `abc` => 0), and a round-trip. The "0 bytes = no limit / empty input" cases are exactly the edge conditions the feature hinges on, and they're real assertions. The page.test.tsx updates correctly switch from asserting `"1 GB"` text to asserting a `spinbutton` with value `"1"`, matching the new editable control. Good.

**upload-size API e2e — good.** `admin settings expose and accept a new max upload size` does GET, asserts `max_upload_size_bytes` present, POSTs a changed value, GETs again and asserts `after.max_upload_size_bytes === target`. Read-after-write is a genuine round-trip assertion, with idempotent restore. No skip.

**The soft spots — `saving a changed upload size issues a settings POST` and the age-policy save test.** Both use the pattern:
```
const postResponse = page.waitForResponse(...).catch(() => null);
await save();
const posted = await postResponse;
if (posted) { expect(posted.status()).toBeLessThan(400); }
```
If the POST never fires (button mis-wired, validation silently blocks, the row's Save targets the wrong mutation), `waitForResponse` times out, `.catch` yields null, the `if (posted)` block is skipped, and the test passes green. That is an assertion that cannot fail on the exact regression it purports to guard. The `body !== 'Application error'` check is too weak to compensate. To actually prove the save: drop the `.catch(()=>null)`/`if` and assert the POST fires AND its body contains the changed value, or assert read-after-write through the UI. As written, the "issues a settings POST" test does not reliably prove a POST is issued.

**age-policy — mixed.** The API tests are real (PATCH stores policy => ok; negative cooldown `-10` => >=400, a genuine negative case). The UI test has good intent (`cooldown` disabled until toggle enabled, then `toBeEnabled` after click) and those ARE distinguishing assertions, but the final save reuses the same conditional `if (saved)` pattern, so the save outcome itself isn't guaranteed asserted. Tighten the save assertions; the disabled/enabled toggle and API tests are fine. Not requesting changes, but the conditional-POST pattern should be fixed before treating these as save-flow coverage.

brandonrc (commented):
Frontend/React correctness review. One real correctness bug (stale initial state); treat it as a blocker even though this is posted as a comment (the review tool won't let me request-changes on my own PR).

### Blocker: `UploadSizeSetting` shows the wrong current value after load

`src/app/(app)/(admin)/settings/page.tsx`

```tsx
const initial = bytesToUploadSize(currentBytes ?? 0);
const [value, setValue] = useState(initial.value);
const [unit, setUnit] = useState<UploadSizeUnit>(initial.unit);
```

The parent `SettingsPage` renders the Storage tab (and therefore this component) unconditionally; it does not gate the whole tab on `settingsLoading` (only individual `SettingRow` values via `storageValue`). So `UploadSizeSetting` mounts while `currentBytes` is still `undefined`, `useState` captures `value = ""`, and `useState` never re-runs when the query resolves. Result: once loading finishes, the editable input shows an empty field ("No limit" placeholder) even when a real limit is configured, misrepresenting the persisted value. Save is disabled until `dirty`, so it won't silently zero the limit, but the displayed state is wrong.

Fix options:
- Render the editor only once data is present (return the loading row until `currentBytes != null`, then mount the editor so its initial state is correct), or
- Add a `key={currentBytes}` to remount on value change, or
- Sync with the "adjust state during render" pattern (track last-seen `currentBytes` in state and reset `value`/`unit` when it changes), guarding against clobbering in-progress edits.

A unit/RTL test that mounts with `loading=true` then flips to `loading=false` with a non-zero `currentBytes` would have caught this.

### Suggestions

- **Age policy form is write-only with fixed defaults.** `ageEnabled`/`ageValue`/`ageUnit` default to `false`/`"3"`/`"days"` and are never seeded from the repo's stored `quarantine_enabled`/`quarantine_duration_minutes`. A repo already quarantining at 7 days will display "disabled / 3 days", and saving overwrites the real config. This is the same write-through tradeoff as #462's release target; if there's no GET to read current state, at least say so in the UI (e.g. "current value not shown") so an admin doesn't assume the form reflects what's stored.
- `settingsApi.updateMaxUploadSize` does a read-modify-write that spreads the entire `getSettings()` response back into the POST. This is the documented approach for the replace-whole-object endpoint, but it (a) has a TOCTOU window if two admins save concurrently and (b) round-trips any read-only/computed fields the GET returns. Acceptable given the endpoint shape; worth a comment that the GET and POST schemas are assumed identical.

The `apiFetch` usage and `bytesToUploadSize`/`uploadSizeToBytes`/`ageToMinutes` helpers are clean and well unit-tested.

### Merge sequencing (not a defect)
Overlaps #462 on `repo-settings-tab.tsx` and #463/others on `settings/page.tsx`. Known, will be merge-sequenced.

brandonrc:

PR #465, opened 2026-06-02
feat: system config feature flags and rate-limit exemption admin UI

### LINKED ISSUE #270: Explore: admin UI for rate limit exemption management
Backend: artifact-keeper/artifact-keeper#680

Rate limit exemptions are currently configured via env vars (RATE_LIMIT_EXEMPT_USERNAMES, RATE_LIMIT_EXEMPT_SERVICE_ACCOUNTS). Explore whether admins would benefit from:

1. Viewing current rate limit configuration in admin settings
2. Seeing which users/service accounts are exempt
3. Managing exemptions via the UI instead of env vars
4. Viewing rate limit metrics (who is being throttled)

Low priority. The env var approach works for now but a UI would improve operator experience.

### LINKED ISSUE #271: Explore: consume /api/v1/system/config for feature flags
Backend: artifact-keeper/artifact-keeper#496

The backend now exposes GET /api/v1/system/config (no auth required) with runtime configuration:
- max_upload_size_bytes, demo_mode
- scanners (trivy, openscap, dependency_track enabled flags)
- search_engine, storage_backend
- auth (oidc_enabled, ldap_enabled, sso_enabled), oidc_issuer

Explore using this endpoint to:
1. Conditionally show/hide UI features based on what's enabled (e.g., hide scanner tab if no scanners configured)
2. Show upload size limits on upload dialogs
3. Show storage backend type in admin dashboard
4. Adapt login page based on SSO availability

This could replace any hardcoded feature detection the frontend currently does.

brandonrc:
## Summary

Two v1.2.0 admin/config features.

**System config feature flags (#271).** New `SystemConfigProvider` fetches `GET /api/v1/system/config` once per session and exposes the result plus derived feature flags via `useSystemConfig` / `useFeatureFlags`. Wired into the obvious gated surfaces:
- Sidebar Security group: "Scan Results" is hidden unless Trivy or OpenSCAP is configured; "DT Projects" is hidden unless the Dependency-Track integration is enabled. The rest of the Security group (policies, permissions, quality gates) always renders.
- Artifact upload dropzone shows the configured max upload size and rejects oversize files client-side before any request is sent.

The provider falls back to permissive defaults on fetch error so a transient outage never hides a feature the operator actually configured (scanner-gated surfaces are the deliberate exception and default to hidden).

**Rate-limit exemption admin UI (#270).** New `/rate-limits` admin page that shows the effective per-window rate limits and lets admins view, add, and remove exemptions for usernames, service accounts, and CIDR ranges (with client-side CIDR validation). Env-sourced exemptions are shown read-only. The page degrades gracefully when the backend has not yet shipped the exemption-management endpoints (config card and exemptions section each render an "unavailable" state instead of crashing).

API clients use the existing `apiFetch` wrapper with zod validation at the trust boundary, matching the pattern in `src/lib/api/settings.ts`.

## Test Checklist
- [x] Manually tested locally

## UI Changes
- [ ] Dark mode verified

E2E specs live under `e2e/suites/interactions/admin/` (`system-config-feature-flags.spec.ts`, `rate-limits.spec.ts`) and run in the `interactions` project against the 1.2.0 backend. Unit suite passing locally (`vitest run`). eslint clean, `tsc --noEmit` introduces no new errors.

Closes #271
Closes #270

### DIFF
--- a/e2e/suites/interactions/admin/rate-limits.spec.ts
+++ b/e2e/suites/interactions/admin/rate-limits.spec.ts
@@ -1,0 +1,120 @@
+import { test, expect } from '@playwright/test';
+
+/**
+ * Admin UI for rate-limit exemption management (issue #270).
+ *
+ * The page lives at /rate-limits and lets an admin view the effective rate
+ * limits and add/remove exemptions for users, service accounts, and CIDR
+ * ranges. The backend exemption-management endpoints may not be present on
+ * every server, so the page is designed to degrade: the config card and the
+ * exemptions section both render an "unavailable" state instead of crashing.
+ * These tests assert the UI surface and the add/remove flow, tolerating a
+ * backend that has not shipped the endpoints yet.
+ */
+test.describe('Rate limit admin', () => {
+  const consoleErrors: string[] = [];
+
+  test.beforeEach(async ({ page }) => {
+    consoleErrors.length = 0;
+    page.on('console', (msg) => {
+      if (msg.type() === 'error') consoleErrors.push(msg.text());
+    });
+    await page.goto('/rate-limits');
+    await page.waitForLoadState('domcontentloaded');
+  });
+
+  test('page loads with Rate Limits heading', async ({ page }) => {
+    await expect(
+      page.getByRole('heading', { name: /rate limits/i }).first()
+    ).toBeVisible({ timeout: 15000 });
+  });
+
+  test('shows the current rate limits and exemptions sections', async ({ page }) => {
+    await expect(
+      page.getByText(/current rate limits/i).first()
+    ).toBeVisible({ timeout: 10000 });
+    await expect(
+      page.getByText(/exemptions/i).first()
+    ).toBeVisible({ timeout: 10000 });
+  });
+
+  test('Add Exemption button opens a dialog with type, value, and note fields', async ({ page }) => {
+    const addBtn = page.getByRole('button', { name: /add exemption/i }).first();
+    await expect(addBtn).toBeVisible({ timeout: 10000 });
+    await addBtn.click();
+
+    const dialog = page.getByRole('dialog');
+    await expect(dialog).toBeVisible({ timeout: 10000 });
+
+    // Type selector (combobox) plus value and note inputs.
+    await expect(dialog.getByRole('combobox').first()).toBeVisible();
+    await expect(dialog.locator('#exemption-value')).toBeVisible();
+    await expect(dialog.locator('#exemption-note')).toBeVisible();
+
+    // Cancel closes the dialog without creating anything.
+    await dialog.getByRole('button', { name: /cancel/i }).click();
+    await expect(dialog).toBeHidden({ timeout: 5000 });
+  });
+
+  test('invalid CIDR is rejected client-side', async ({ page }) => {
+    await page.getByRole('button', { name: /add exemption/i }).first().click();
+    const dialog = page.getByRole('dialog');
+    await expect(dialog).toBeVisible({ timeout: 10000 });
+
+    // Switch the type to CIDR.
+    await dialog.getByRole('combobox').first().click();
+    const cidrOption = page.getByRole('option', { name: /cidr/i });
+    await cidrOption.click();
+
+    await dialog.locator('#exemption-value').fill('not-a-valid-cidr');
+    await dialog.getByRole('button', { name: /^add exemption$/i }).click();
+
+    // A validation toast appears and the dialog stays open.
+    await expect(page.getByText(/valid cidr/i).first()).toBeVisible({ timeout: 8000 });
+    await expect(dialog).toBeVisible();
+
+    await dialog.getByRole('button', { name: /cancel/i }).click();
+  });
+
+  test('add then remove a username exemption (when the backend supports it)', async ({ page }) => {
+    const username = `e2e-exempt-${Date.now()}`;
+
+    await page.getByRole('button', { name: /add exemption/i }).first().click();
+    const dialog = page.getByRole('dialog');
+    await expect(dialog).toBeVisible({ timeout: 10000 });
+
+    // Default type is username; just fill the value and submit.
+    await dialog.locator('#exemption-value').fill(username);
+    await dialog.locator('#exemption-note').fill('e2e test exemption');
+    await dialog.getByRole('button', { name: /^add exemption$/i }).click();
+
+    // Either the backend accepts it (row appears) or it has no exemption
+    // endpoint (failure toast). Both are acceptable, but a success must round
+    // trip to a removable row.
+    const newRow = page.getByRole('row', { name: new RegExp(username) });
+    const created = await newRow
+      .isVisible({ timeout: 8000 })
+      .catch(() => false);
+
+    if (!created) {
+      test.skip(true, 'Backend does not expose rate-limit exemption management');
+      return;
+    }
+
+    // Remove it and confirm in the alert dialog.
+    await newRow.getByRole('button', { name: new RegExp(`remove exemption ${username}`, 'i') }).click();
+    const confirm = page.getByRole('alertdialog');
+    await expect(confirm).toBeVisible({ timeout: 8000 });
+    await confirm.getByRole('button', { name: /^remove$/i }).click();
+
+    await expect(newRow).toHaveCount(0, { timeout: 10000 });
+  });
+
+  test('no uncaught console errors on load', async ({ page }) => {
+    await page.waitForTimeout(1500);
+    const fatal = consoleErrors.filter(
+      (e) => !/favicon|hydrat|ResizeObserver/i.test(e)
+    );
+    expect(fatal, fatal.join('\n')).toHaveLength(0);
+  });
+});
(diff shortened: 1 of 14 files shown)


brandonrc (commented):
Accessibility review (WCAG 2.2 AA) of the rate-limit admin page (table, add/remove dialogs) and feature-flag-gated nav.

**Positive findings**
- Add Exemption dialog: all three controls are labelled. `<Label htmlFor="exemption-type">` + `SelectTrigger id`, `<Label htmlFor="exemption-value">` (label text is dynamic via `meta.label`, so the field name tracks the selected type), `<Label htmlFor="exemption-note">`. Good (4.1.2, 3.3.2).
- Remove flow uses `AlertDialog` (Radix) with `AlertDialogTitle`/`AlertDialogDescription`: proper modal alertdialog with focus trap and Escape. Good (2.4.3, 4.1.2).
- Per-row remove button has `aria-label={`Remove exemption ${ex.value}`}` (icon-only button with an accessible name) plus `title`; the env-sourced disabled state is native (4.1.2, 4.1.1).
- ConfigCard uses a real `<dl>/<dt>/<dd>` for read-only config (1.3.1). Nice.
- Feature-flag nav gating filters items from the array before render, so hidden entries are removed from the DOM (not just visually hidden) and are unreachable by keyboard/SR. Correct approach, no phantom focus targets.

**Gaps**

1. Serious (3.3.1 / 4.1.3): the Add Exemption form reports validation only via `toast.error(validateExemption(...))`. The error is transient and not tied to the offending field. On invalid input, `#exemption-value` should set `aria-invalid` and `aria-describedby` to a persistent `role="alert"` message inside the dialog, rather than relying on a toast that can render outside the dialog and is not associated with the control. Same recurring pattern as the rest of the v1.2.0 batch (see #459's create-dialog pattern).

2. Moderate (1.3.1): the exemptions table has an empty action header `<TableHead className="w-[1%]" />`. Add `<span className="sr-only">Actions</span>` so the column is named for SR. Also add `aria-label="Rate-limit exemptions"` on the `<Table>` to match the named-table pattern used in #462.

3. Minor (1.4.1): Source is conveyed by Badge variant (Environment vs Manual), but the text labels carry the meaning, so this is fine.

4. Minor: the "—" empty-note placeholder is read as "em dash" by SR; low priority, consider sr-only "no note".

No hard blocker. Item 1 (dialog form errors not programmatically associated) is the one to fix for consistency. Review-only, not approving.

brandonrc (commented):
Security/API review (CIDR validation + sensitive-config exposure lens).

Net: safe. Admin surface is gated, the public config endpoint exposes only booleans/limits, and client CIDR validation is advisory with the backend as the authority.

Authorization — OK. The page lives under `src/app/(app)/(admin)/rate-limits/page.tsx`, and the `(admin)` route group is wrapped by `RequireAdmin` in its layout, so non-admins cannot render it. The exemption endpoints are documented as admin-guarded server-side (`/api/v1/admin/rate-limits[/exemptions]`). Client gating is UX only; the real check must stay on the backend (it is, per the admin router) — confirm that holds, since a non-admin could otherwise call these directly.

isValidCidr / validateExemption — ADVISORY, ACCEPTABLE. The comment correctly states the backend does authoritative validation. The client check is a reasonable sanity filter: IPv4 requires 4 octets 0-255 and prefix <=32; IPv6 requires prefix <=128 and a `[0-9a-fA-F:]` charset. It is intentionally loose (e.g. it would accept malformed IPv6 like `:::/64` or out-of-range-looking forms), but since it only gates the submit button and the backend is authoritative, this is not a security issue — just do not rely on it as a trust boundary. No ReDoS: the regexes are linear, no catastrophic backtracking.

Sensitive config exposure — OK. systemConfigApi consumes `GET /api/v1/system/config`, which returns only non-sensitive booleans/strings (scanner/auth enable flags, search/storage backend names, upload limit, demo/guest flags, optional `oidc_issuer`). No secrets, connection strings, CIDRs, usernames, or token material are surfaced to non-admins. `oidc_issuer` is a public URL by design. zod `.passthrough()` with explicit field mapping means a future backend field that accidentally carries something sensitive would still need the web app to model+render it, so no silent leak through this client. Defaults are permissive for nav visibility but reveal nothing.

Note (out of scope, confirm server-side): ensure `/api/v1/system/config` truly returns only these flags and never echoes the raw env values (e.g. the actual `RATE_LIMIT_TRUSTED_CIDRS` list) to unauthenticated callers. The web type does not request them; just verify the backend response does not include them.

No request-changes.

brandonrc (commented):
Test/E2E quality review (v1.2.0 batch). Verdict: system-config spec is genuinely strong; rate-limits spec leans on a skip and tolerant branches that can pass without testing the add/remove flow.

**system-config (#271) — strong.** `public system config endpoint returns the expected shape` asserts the full contract: typed checks on `max_upload_size_bytes`, `demo_mode`, `guest_access_enabled`, `search_engine`, `storage_backend`, plus nested `scanners.*` and `auth.*` booleans. No skip. The `scanner nav entries match the backend scanner flags` test is the best part: it reads the live config and asserts BOTH directions — if scanners on, the nav link `toBeVisible`; else `toHaveCount(0)`. That two-sided assertion genuinely proves the gating logic rather than just "link exists." This distinguishes fixed-vs-broken regardless of how the CI backend is configured.

**upload-size advertise test.** Has two skips (no limit advertised; upload tab unavailable) then asserts `getByText(/max\s/i)`. The `/max\s/i` matcher is loose — it would match almost any "max ..." string on the page, not necessarily the configured size. Weak assertion behind two skips; low value but not harmful.

**rate-limits (#270) — the soft spots.**
- `add then remove a username exemption` is the core round-trip, and it's gated: if the new row isn't visible in 8s, `test.skip(true, 'Backend does not expose rate-limit exemption management')`. So on any backend without the endpoint (or if creation is slow), the entire add/remove/confirm flow never asserts. The remove path (`toHaveCount(0)`) only runs in the happy case. This is the load-bearing behavioral test and it can no-op.
- `invalid CIDR is rejected client-side` is good and non-skippable: switches type to CIDR, fills `not-a-valid-cidr`, asserts a `/valid cidr/i` toast AND `dialog.toBeVisible()` (stays open). Real negative case proving client validation. Keep this.
- The dialog-fields and heading tests are presence smoke checks, fine as such.
- `no uncaught console errors on load` filters favicon/hydrate/ResizeObserver — reasonable, but it's a guard, not coverage of the feature.

**Recommendation.** For #270 to have non-conditional coverage of the actual exemption management, either confirm the v1.2.0 backend ships the endpoints (then drop the skip so a broken add/remove fails), or mock the endpoint. As-is, the CIDR-validation and the system-config nav-gating tests are the genuine coverage; the exemption round-trip is conditional. Not requesting changes — system-config carries this and the CIDR test is real.

brandonrc (commented):
Frontend/React correctness review. No blockers. The strongest PR in the batch: clean provider, zod at the trust boundary, good error/unavailable states, and complete tests.

### Correctness notes

- **`SystemConfigProvider` placement above `AuthProvider`** is correct given the endpoint is public (`/api/v1/system/config`, documented "requires no authentication"). The query is unconditional with `staleTime: 10m` and `retry: false`, which is appropriate for a public, rarely-changing config. Because it's public, there's no need to refetch on login. Good.

- **Permissive-defaults design is sound.** `DEFAULT_SYSTEM_CONFIG` keeps nav surfaces visible on fetch failure, with the deliberate scanner exception defaulting to hidden, and the rationale is documented. On error/loading the context still exposes a concrete `SystemConfig`, so consumers never null-check. The provider throws a clear error when used outside its scope, with a test.

- **Sidebar gating** (`visibleSecurityItems`) keys off the real hrefs (`/security/scans`, `/security/dt-projects`, verified against `securityItems`), gates Scan Results on Trivy-or-OpenSCAP and DT Projects on Dependency-Track, and leaves the rest of the Security group always visible. Matches the stated intent and is unit-tested both ways.

- **`file-upload.tsx` client-side size check** is correct defense-in-depth: it rejects oversize files before upload (clears the input, sets an error, returns early) while the backend still enforces authoritatively. `handleFile` deps correctly include `maxUploadSizeBytes`. The advertised max in the dropzone copy only shows when > 0.

- **Rate-limits page** has proper admin guard, distinct loading/error/empty/unavailable states, a confirm dialog for the destructive remove, and env-sourced rows shown read-only (delete disabled with an explanatory title). Mutations invalidate the right query keys. This is the model the other admin pages (e.g. #462 routing rules) should follow for error UX.

- **API clients** (`system-config.ts`, `rate-limits.ts`) validate every response with zod `.safeParse` + `.passthrough()` for forward-compat, accept both array and `{exemptions: [...]}` list shapes, and `encodeURIComponent` the delete id. The `apiFetch` escape hatch is used as intended.

### Nits
- `isValidCidr`: an empty prefix like `10.0.0.0/` parses as `Number("") === 0`, a valid integer >= 0, so it passes as `/0`. Backend validates authoritatively, but tightening to require a non-empty prefix string would catch the obvious typo client-side.
- `rate(window)` returns `${limit} requests` when `window_secs <= 0`; fine, just confirm the backend never sends a 0 window where a real window is expected.

LGTM.

brandonrc:

PR #509, opened 2026-06-22
feat(promotion-rules): Promotion Rules admin UI

brandonrc:
## Summary

Fourth of the **1.2.1 endpoint build-out**. The backend promotion-rules API (`/api/v1/promotion-rules*`) had no web UI; this adds a **Promotion Rules** admin page (Operations nav).

### `/promotion-rules` (admin)
- List rules — `source → target` repo (keys resolved), auto-promote / signed / disabled badges, gate summary
- **Create / edit** gating criteria: max CVE severity, min health score, min staging hours, require-signature, auto-promote, enabled
- **source/target repos are immutable** after create — the SDK's `UpdateRuleRequest` omits them, so the edit form locks them to read-only (and the update body never sends them)
- **Evaluate** (dry-run) a rule → `passed/total, failed` toast
- Delete with confirm; loading / empty / error / non-admin states

### Code
- `src/lib/api/promotion-rules.ts` — typed adapter over `listRules`/`getRule`/`createRule`/`updateRule`/`deleteRule`/`evaluateRule`. Uses `PromotionRuleResponse` — its **native, consistent** type (unlike the curation rules' mismatched `RuleResponse`, which is why those were deferred).
- "Promotion Rules" entry in the Operations nav + barrel export.

## Tests
- API adapter: 8 cases (list mapping incl. `null→[]`, create/update bodies, the no-source/target update, `evaluate` mapping `total_artifacts→total`, errors).
- Page: 9 cases — admin gate, empty/loading/error, list with resolved repo keys, create (name+source+target), edit locks source/target, evaluate, delete, and mutation callbacks verifying update omits source/target.
- `npm run build` ✅ (`/promotion-rules` route), all tests green.

## Scope note
`allowed_licenses` (array) and `max_artifact_age_days` are in the adapter but not yet surfaced in the form — follow-up.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/app/(app)/(admin)/promotion-rules/page.test.tsx
+++ b/src/app/(app)/(admin)/promotion-rules/page.test.tsx
@@ -1,0 +1,240 @@
+// @vitest-environment jsdom
+import React from "react";
+import { describe, it, expect, vi, beforeAll, beforeEach, afterEach } from "vitest";
+import "@testing-library/jest-dom/vitest";
+import { render, screen, cleanup, within } from "@testing-library/react";
+import userEvent from "@testing-library/user-event";
+
+beforeAll(() => {
+  globalThis.ResizeObserver = class {
+    observe() {}
+    unobserve() {}
+    disconnect() {}
+  };
+  Element.prototype.scrollIntoView = vi.fn();
+  Element.prototype.hasPointerCapture = vi.fn();
+  Element.prototype.releasePointerCapture = vi.fn();
+});
+
+interface MutationConfig {
+  mutationFn: (...a: unknown[]) => unknown;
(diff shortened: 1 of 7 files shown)


brandonrc:
Addressed the review **[blocker]**: `updateRule` is a PUT (replace), so a name-only edit was silently dropping `allowed_licenses` and `max_artifact_age_days` (wiping a license allow-list / age gate). Both fields are now surfaced in the form (a comma-separated licenses input + a max-age number field), read back in `openEdit`, and included in the update body — so edits round-trip the full object. Added a regression test that loads a rule with those gates set, does a name-only edit, and asserts both survive the update. Also extended the create/update callback test to assert the gates are sent.

Issue #406, opened 2026-05-11
Changing OIDC settings in the frontend overwrites OIDC settings that are already in the db


Brotholomew:
### Description

1. The frontend does not allow to change all OIDC settings: e.g. redirect_uri
2. When changing the settings of the OIDC provider that is already configured, the settings that are not in the frontend get overwritten - the `attribute_mapping` in the database gets fully overwritten with values from the frontend (ignoring the previously set redirect_uri

### Environment Information

- Browser: this is browser independent - tested in MS Edge
- Browser Version: MS Edge v147.0.3912.98
- Operating System: Windows 11 Enterprise 24H2
- Screen Resolution: 2550x1440
- Connection type: the whole artifact-keeper ecosystem deployed on a kube cluster behind an ingress controller; no additional proxies between my host and the frontend deployment

### Web UI Version

1.1.3

### Backend Version

1.1.2

### Expected Behavior

Only the `name` column in  the db get's updated (or at least the `attribute_mapping` does not get overwritten but updated).

### Observed Behavior

The `name` and the `attribute_mapping` get's changed

### Steps to Reproduce

1. Run an artifact-keeper instance with OIDC settings via env variables:
- OIDC_ISSUER
- OIDC_CLIENT_ID
- OIDC_CLIENT_SECRET
- OIDC_EMAIL_CLAIM
- OIDC_GROUPS_CLAIM
- OIDC_SCOPES
- OIDC_REDIRECT_URI
- ALLOW_LOCAL_ADMIN_LOGIN
- OIDC_ADMIN_GROUP
2. Change the 'name' of the OIDC provider in the UI settings 

### Browser Console Logs

```shell

```

### Pre-submission Checklist

- [x] I have searched existing issues to make sure this is not a duplicate.
- [x] I have redacted any sensitive information (tokens, passwords, internal URLs).
- [x] I have tested this on the latest version of the web interface.

PR #523, opened 2026-06-26
fix(sso): surface SAML use_absolute_acs_url opt-in on the provider edit form

### LINKED ISSUE #521: Expose the new SAML use_absolute_acs_url opt-in on the provider edit form
### Problem Statement

Companion to artifact-keeper backend issue artifact-keeper/artifact-keeper#2038 ("SAML AuthnRequest emits a relative AssertionConsumerServiceURL, rejected by stricter IdPs"). Once the backend lands the per-provider `use_absolute_acs_url` flag (migration 138), the SSO admin UI needs to surface it on the SAML provider edit dialog so operators of strict-IdP deployments can flip it without poking at the database. Without UI coverage, the only way to set the flag is a direct `PUT /api/v1/admin/sso/saml/<id>` against the OpenAPI — fine for power users but inconsistent with how every other SAML toggle (Sign Requests, Require Signed Assertions, NameID Format) is exposed today.

### Proposed Solution

- Add a "Use absolute ACS URL" Switch to the SAML provider edit dialog (Admin → Settings → SSO → SAML Providers), in the same opt-in group as Sign Requests / Require Signed Assertions. Off by default.
- Help text explains what the IdP receives when the flag is on and identifies the use case ("stricter SAML 2.0 IdPs that reject relative AssertionConsumerServiceURLs").
- Initial state sourced from the loaded config; submit payload echoes the toggle's value back to both create and update endpoints.
- Local types (`src/types/sso.ts`) and the adapter (`src/lib/api/sso.ts::adaptSamlConfig`) propagate the new field with a defensive default so older backends that have not yet shipped migration 138 keep their pre-138 behaviour.

PR to follow this issue.

### Alternatives Considered

Power users can already `PUT` the flag via the OpenAPI; this issue asks for parity with every other SAML provider field in the admin UI.

### Use Case

As an operator integrating Artifact Keeper with a strict SAML IdP (e.g. Lark AnyCross), I want to enable the absolute ACS URL flag from the SSO admin UI on the affected provider, so that I do not need to issue a raw API call to flip a per-provider setting.

### Component

Settings

### Additional Context

Depends on backend migration 138 / artifact-keeper/artifact-keeper#2038. The UI change is gated on the backend regenerating the OpenAPI spec and republishing the TypeScript SDK; the adapter ships a defensive default so the UI is safe to deploy against an older backend (the form simply starts at `false`, which is the pre-138 behaviour).

### Pre-submission Checklist

StringKe:
## Summary

Companion to the artifact-keeper backend change that adds a per-provider `use_absolute_acs_url` flag (tracked in artifact-keeper/artifact-keeper#2038, migration 138). The flag toggles whether the SAML AuthnRequest emits an absolute `AssertionConsumerServiceURL` or the historical relative path; stricter SAML 2.0 IdPs (notably Lark AnyCross and certain enterprise deployments) reject a relative ACS URL outright.

UI changes (Admin → Settings → SSO → SAML Providers, edit dialog):

- New **Use absolute ACS URL** Switch sits below "Require Signed Assertions" in the same opt-in group. Off by default — existing providers keep their pre-138 wire format.
- The switch help text spells out what the IdP receives and why an operator would flip it on so the change is self-explanatory at the edit screen without diving into release notes.
- Initial state on edit is sourced from `config.use_absolute_acs_url`; the submit payload echoes the toggle's value back to the backend on both create and update.

Type / adapter changes:

- `SamlConfig` (response shape) gains `use_absolute_acs_url: boolean`. `CreateSamlConfigRequest` / `UpdateSamlConfigRequest` gain `use_absolute_acs_url?: boolean` so the toggle round-trips. Mirrors the backend's `Option<bool>` convention on the request side and the always-present `bool` on the response side.
- `adaptSamlConfig` defensively reads the field via a narrowed cast and defaults to `false` when absent. The regenerated `@artifact-keeper/sdk` drops in automatically on the next backend release tag (the `sync-openapi-spec` workflow pushes the new schema to `artifact-keeper-api` and that repo's release pipeline republishes the TS SDK); until then, talking to an older backend that doesn't return the field simply means the form starts at `false`, which is exactly the pre-138 behaviour.

## Test Checklist
- [x] Manually tested locally

Two new Vitest cases in `SSO SAML use_absolute_acs_url toggle`:

- `preserves the existing use_absolute_acs_url value on an unrelated edit` — opens the edit dialog on a provider with the flag off, changes only Name, asserts the outbound `updateSaml` payload still carries `use_absolute_acs_url: false`. Pins the no-accidental-flip contract.
- `propagates the toggled value into the update payload` — clicks the new switch, saves, asserts `use_absolute_acs_url: true` reaches the update call.

The existing `SAML_WITH_EXTRA_MAPPING` fixture is extended with `use_absolute_acs_url: false` so the prior #406 attribute_mapping regression test continues to model the realistic response shape.

Verification:

- `npm run lint` clean on changed files (0 errors; the 36 warnings are pre-existing in unrelated files).
- `npm test -- 'src/app/(app)/(admin)/settings/sso'`: 4/4 passed, including the two new cases.

## UI Changes
- [x] Responsive layout verified (mobile + desktop) — the new switch sits inside the existing `flex items-center justify-between` row pattern reused from Sign Requests / Require Signed Assertions, so responsive behaviour matches the surrounding controls.
- [x] Dark mode verified — uses default `text-muted-foreground` for help copy; no custom colours that would diverge between themes.
- [x] Accessibility checked (keyboard navigation, screen reader) — Switch has `id="saml-use-absolute-acs-url"` with the matching Label so screen readers announce the toggle and its help text correctly; tab order follows the surrounding controls.
- [ ] N/A - no UI changes

Playwright spec is intentionally not added here — the dependent backend change has to land first so an end-to-end browser test against a fresh stack can exercise the full opt-in flow. Happy to add one in a follow-up PR once the backend PR is in.

Fixes #521

### DIFF
--- a/src/app/(app)/(admin)/settings/sso/__tests__/page.test.tsx
+++ b/src/app/(app)/(admin)/settings/sso/__tests__/page.test.tsx
@@ -222,6 +222,7 @@ const SAML_WITH_EXTRA_MAPPING = {
   admin_group: "artifact-keeper-admins",
   sign_requests: true,
   require_signed_assertions: true,
+  use_absolute_acs_url: false,
   is_enabled: true,
   created_at: "2025-01-01T00:00:00Z",
   updated_at: "2025-01-01T00:00:00Z",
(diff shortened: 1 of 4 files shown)


brandonrc:
Closing as superseded — the `use_absolute_acs_url` opt-in landed on main via the #521 work (see `src/app/(app)/(admin)/settings/sso/page.tsx` and the `SSO SAML use_absolute_acs_url toggle (#521)` test block). Thanks @StringKe — your PR covered the same ground and confirmed the approach.

PR #524, opened 2026-06-26
feat(sso): surface OIDC allow_legacy_rsa_keys opt-in on the provider edit form

### LINKED ISSUE #522: Expose the new OIDC allow_legacy_rsa_keys opt-in on the provider edit form
### Problem Statement

Companion to artifact-keeper backend issue artifact-keeper/artifact-keeper#2039 ("Opt-in compatibility for OIDC IdPs that sign ID tokens with sub-2048-bit RSA keys"). Once the backend lands the per-provider `allow_legacy_rsa_keys` flag (migration 139), the SSO admin UI needs to surface it on the OIDC provider edit dialog so operators stuck on legacy IdPs (most prominently Lark AnyCross with its 1024-bit RS256 key) can flip the compat flag from the admin UI rather than the API. The toggle also wants explicit on-screen warning copy because the operator is intentionally accepting a key strength below the OWASP ASVS 4.0 baseline; the security caveat must be visible at the point of decision, not buried in release notes.

### Proposed Solution

- Add an "Allow legacy RSA keys (< 2048-bit)" Switch to the OIDC provider edit dialog (Admin → Settings → SSO → OIDC Providers), in the same opt-in group as Auto Create Users. Off by default.
- Help text spells out what gets accepted when the flag is on (sub-2048-bit RSA via RS256/384/512 PKCS#1 v1.5 fallback), why an operator would enable it (Lark AnyCross is named directly), and the trade-off — an amber-coloured "Below the OWASP ASVS 4.0 baseline" line keeps the security caveat in front of the operator.
- Initial state sourced from the loaded config; submit payload echoes the toggle's value back to both create and update endpoints.
- Local types (`src/types/sso.ts`) and the adapter (`src/lib/api/sso.ts::adaptOidcConfig`) propagate the new field with a defensive default so older backends that have not yet shipped migration 139 keep their pre-139 strict-verification behaviour.

PR to follow this issue.

### Alternatives Considered

Same as the backend: leaving it API-only is fine for power users but inconsistent with every other OIDC provider field in the admin UI, and the security caveat would not be visible at the point of decision.

### Use Case

As an operator rolling out Lark AnyCross OIDC for an internal team, I want a one-click toggle on the affected provider with a visible OWASP-baseline warning, so that I can knowingly accept the security trade-off scoped to that single provider without dropping the strict path for every other OIDC IdP in the deployment.

### Component

Settings

### Additional Context

Depends on backend migration 139 / artifact-keeper/artifact-keeper#2039. The adapter ships a defensive default so the UI is safe to deploy against an older backend (the form starts at `false`, which is the pre-139 strict behaviour).

### Pre-submission Checklist

StringKe:
## Summary

Companion to the artifact-keeper backend change that adds a per-provider `allow_legacy_rsa_keys` flag (tracked in artifact-keeper/artifact-keeper#2039, migration 139). The flag opts into a restricted manual `rsa` + `sha2` PKCS#1 v1.5 fallback that accepts ID tokens signed with RSA keys shorter than 2048 bits, scoped to RS256 / RS384 / RS512 only. The strict path keeps enforcing the OWASP ASVS 4.0 V6.2.5 baseline by default; this UI lets an operator opt a specific provider out — typically for legacy IdPs such as Lark AnyCross.

UI changes (Admin → Settings → SSO → OIDC Providers, edit dialog):

- New **Allow legacy RSA keys (< 2048-bit)** Switch sits below "Auto Create Users" in the same opt-in group. Off by default — existing providers keep their pre-139 strict-verification behaviour.
- Help text spells out what gets accepted when the flag is on (sub-2048-bit RSA via RS256/384/512 PKCS#1 v1.5 fallback), why an operator would enable it (Lark AnyCross is named directly), and the trade-off — an amber-coloured "Below the OWASP ASVS 4.0 baseline" line ensures reviewers and operators land on the security caveat at the point of decision rather than in a release-notes afterthought.
- Initial state on edit is sourced from `config.allow_legacy_rsa_keys`; the submit payload echoes the toggle's value back to the backend on both create and update.

Type / adapter changes:

- `OidcConfig` (response shape) gains `allow_legacy_rsa_keys: boolean`. `CreateOidcConfigRequest` / `UpdateOidcConfigRequest` gain `allow_legacy_rsa_keys?: boolean` so the toggle round-trips on both paths.
- `adaptOidcConfig` defensively reads the field via a narrowed cast and defaults to `false` when absent — matches the migration's default and preserves the strict pre-139 behaviour when talking to an older backend whose regenerated SDK has not yet shipped the field.

## Test Checklist
- [x] Manually tested locally

Three new Vitest cases in `SSO OIDC allow_legacy_rsa_keys toggle`:

- `preserves the existing allow_legacy_rsa_keys value on an unrelated edit` — opens the edit dialog on a provider with the flag off, changes only Name, asserts the outbound `updateOidc` payload still carries `allow_legacy_rsa_keys: false`. Pins the no-accidental-flip contract.
- `propagates the toggled value into the update payload` — clicks the new switch, saves, asserts `allow_legacy_rsa_keys: true` reaches the update call.
- `surfaces the OWASP-baseline warning in the toggle help text` — asserts the amber-coloured warning is rendered. Pins the wording so a future refactor cannot quietly soften it.

The existing `OIDC_WITH_EXTRA_MAPPING` fixture is extended with `allow_legacy_rsa_keys: false` so the prior #406 attribute_mapping regression test continues to model the realistic response shape.

Verification:

- `npm run lint` clean on changed files (0 errors; the 36 warnings are pre-existing in unrelated files).
- `npm test -- 'src/app/(app)/(admin)/settings/sso'`: 5/5 passed (2 prior + 3 new).

## UI Changes
- [x] Responsive layout verified — same `flex items-center justify-between` row pattern as the surrounding toggles (Auto Create Users, etc.).
- [x] Dark mode verified — `text-muted-foreground` for the help copy and `text-amber-600 dark:text-amber-400` for the warning line, which has explicit dark-mode tuning so the caveat reads clearly in both themes.
- [x] Accessibility checked (keyboard navigation, screen reader) — Switch has `id="oidc-allow-legacy-rsa-keys"` with the matching Label; the warning line is in the same DOM block as the Label so it is read out alongside the toggle's purpose.
- [ ] N/A - no UI changes

Playwright spec is intentionally not added here — the dependent backend change has to land first so an end-to-end browser test against a fresh stack can exercise the legacy-RSA verification behaviour. Happy to add one in a follow-up PR once the backend PR is in.

Fixes #522

### DIFF
--- a/src/app/(app)/(admin)/settings/sso/__tests__/page.test.tsx
+++ b/src/app/(app)/(admin)/settings/sso/__tests__/page.test.tsx
@@ -191,6 +191,7 @@ const OIDC_WITH_EXTRA_MAPPING = {
   },
   auto_create_users: true,
   map_groups_to_groups: false,
+  allow_legacy_rsa_keys: false,
   is_enabled: true,
   created_at: "2025-01-01T00:00:00Z",
   updated_at: "2025-01-01T00:00:00Z",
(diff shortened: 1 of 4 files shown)


brandonrc:
Rebased onto current main and resolved the conflicts (main had since landed `map_groups_to_groups` and the SAML `use_absolute_acs_url` toggle; both kept, this toggle added alongside). Note: the backend field shipped in migration **144** (`backend/migrations/144_oidc_allow_legacy_rsa_keys.sql`), not 139 as the commit message says — code comments were corrected during rebase. SSO unit tests: 8/8 pass.

brandonrc (approved):

Issue #254, opened 2026-04-12
Add Maven component grouping toggle to artifact browser


brandonrc:
Backend PR artifact-keeper/artifact-keeper#701 added a `?group_by=maven_component` query parameter to the artifact listing endpoint. For Maven/Gradle repositories, the frontend should:

1. Add a toggle/button to switch between flat artifact list and grouped-by-component view
2. In grouped view, show GAV (groupId:artifactId:version) hierarchy with expandable file lists
3. Only show the toggle for Maven/Gradle format repositories

API: GET `/api/v1/repositories/:key/artifacts?group_by=maven_component` returns a `components` array alongside the regular `items` array.

PR #561, opened 2026-07-09
fix(repositories): use server-side docker_tag grouping in the Docker tag view

### LINKED ISSUE #559: Docker Grouped view builds tags client-side from one page of the flat artifact list — shows "No image tags found" whenever manifests sort past the first page
### Description

The repository page's Docker **Grouped** view does not use the server's `group_by=docker_tag` endpoint (which has existed and worked since artifact-keeper/artifact-keeper#1336). Instead it fetches the **flat** artifact list (`per_page=500`, which the backend clamps to 100 in `list_artifacts`) sorted `ORDER BY path`, and re-derives tags client-side by parsing each row's path for `/manifests/` / `/blobs/` (`groupDockerArtifacts` in `_lib/docker-grouping.ts`).

Any Docker repository where the first 100 rows by path contain no `v2/<image>/manifests/<tag>` rows renders **"No image tags found"** even though tags exist and the server-side grouped query returns them. This is trivially hit by:

- (a) repositories with migrated/legacy rows whose paths sort before `v2/`, and
- (b) **any large native repository** — e.g. images named `aaa*` contribute `v2/aaa/blobs/…` rows that fill page 1 before `v2/zzz/manifests/…` is ever reached.

This is exactly the failure mode predicted in artifact-keeper/artifact-keeper#1194 when the server rollup was requested ("Large Docker repos (multi-tag, multi-arch) will exceed this and silently miss tags", "Pattern-matching is brittle"). The backend endpoint was merged in artifact-keeper/artifact-keeper#1336 but the frontend was never switched over — there is no `docker_tag` string anywhere in the built web chunks.

Side effects of the client-side approach beyond the empty view:
- "Total size" shows the manifest document size, not the real image size (the rollup returns a multi-arch-aware total).
- The 500-row cap silently truncates grouping input even when page 1 does contain manifests.

### Steps to Reproduce

1. Create a Docker local repository and push >100 artifact rows whose paths sort before the first manifest row (e.g. many images whose blob rows alphabetically precede another image's `manifests/` rows — pushing ~4 multi-layer images named `aaa/…` before one named `zzz/…` is enough, since every blob row counts).
2. Open the repository page → Artifacts tab → Grouped view.
3. The view shows "No image tags found" even though `GET /api/v1/repositories/{key}/artifacts?group_by=docker_tag` returns all tags.

### Expected Behavior

The Grouped view lists every tag in the repository, with correct sizes, using the server rollup — same pattern as the Maven component grouping (#254 → backend artifact-keeper/artifact-keeper#701).

### Environment Information

- web 1.3.0 / backend 1.3.0
- Kubernetes (Azure AKS), also reproducible on docker-compose
- Verified against the deployed web bundle (`DockerTagList` groups client-side; no `docker_tag` string in the built chunks) and backend `list_artifacts` (`per_page` clamp to 100, `ORDER BY path` in `artifact_service::list`).

### Suggested fix

Make the Docker Grouped view call `GET /api/v1/repositories/{key}/artifacts?group_by=docker_tag` (per-tag digest, multi-arch-aware total size, last push, scan rollup — added in artifact-keeper/artifact-keeper#1336) instead of re-implementing grouping over one page of the flat list, mirroring how the Maven grouped view consumes `group_by=maven_component`. I have a PR ready.

garrettKelleyCyOne:
## Summary

Closes #559

The Docker **Grouped** view re-derived tag rows client-side from ONE page of the flat artifact list (`per_page=500`, clamped to 100 by the backend, `ORDER BY path`). Any repository whose first page contained no `v2/<image>/manifests/<tag>` rows rendered **"No image tags found"** even though tags existed — trivially hit by large registries (early-sorting blob rows fill page 1) and by repos with rows sorting before `v2/`. It also showed manifest-document sizes instead of real image sizes and relied on URL-shape heuristics (all three flaws called out in artifact-keeper/artifact-keeper#1194 when the server rollup was designed).

This PR switches the view to the server-side rollup that has existed since artifact-keeper/artifact-keeper#1336 — `GET /api/v1/repositories/{key}/artifacts?group_by=docker_tag` — following the exact pattern the Maven component grouping already uses (#254):

- `DockerTagList` now renders `DockerTag` rollups: every tag in the repo, true multi-arch-aware total size, last push time, scan rollup status, a `multi-arch` badge for image indexes, and server-side pagination via `DataTablePagination` (like the Maven list, cf. #443).
- `artifactsApi.listGrouped` models `group_by: 'docker_tag'` and passes `docker_tags` through (SDK regen still pending, same as `maven_component`).
- Tag click resolves the manifest artifact by its deterministic `v2/{image}/manifests/{tag}` path — same mechanism as Maven grouped file clicks (#444); the scan action uses the artifact id the rollup carries.
- The client-side grouping helpers (`groupDockerArtifacts`, `parseDockerPath`) and the "Show layers" disclosure are removed — raw blobs remain inspectable via Flat view; only `truncateDigest` remains.

Net: −360 lines.

## Test Checklist
- [x] Unit tests added/updated — `DockerTagList` suite rewritten against the rollup shape (loading/empty, truncated digest + copy, multi-arch badge, scan status, pagination, click handlers); API client tests cover `group_by=docker_tag` URL + `docker_tags` passthrough; `docker-grouping` lib tests reduced to `truncateDigest`.
- [x] E2E Playwright tests added/updated — the grouped-view spec's tag-row/digest assertions still apply unchanged; the "Show layers" case is removed with the feature.
- [x] Manually tested locally — against a backend 1.3.0 registry with >100 artifact rows before the first manifest row: grouped view now lists all tags (previously "No image tags found"), sizes match the rollup, pagination works.
- [x] No regressions in existing tests — full vitest suite green (2525 tests), `next build` clean.

## UI Changes
- [x] Playwright E2E spec covers the change (`artifact-browser-grouping.spec.ts` Docker grouped-view case)
- [x] Responsive layout verified (mobile + desktop) — table keeps its `overflow-x-auto` wrapper; no layout change beyond the new badge/pagination
- [x] Dark mode verified — only existing themed primitives (Badge, Button, DataTablePagination) are used
- [x] Accessibility checked — aria-labels for row buttons/digests preserved, loading state announced via `role=status`, pagination is the shared accessible component


brandonrc:
Maintainer check-in: the underlying bug (#559 — Docker grouped view shows "No image tags found" when manifests sort past the first page) is still live and this is the right direction (server-side `docker_tag` grouping). The branch now conflicts with main and is ~3 weeks stale — could you rebase? Two heads-ups for the rebase: (1) #692 consolidated repository-list queries into `useRepositories()` in `src/hooks/use-repositories.ts`, and (2) #694 standardized lib/api modules to named exports + `unwrap()` from `src/lib/sdk-utils.ts`. If you no longer have cycles, say the word and a maintainer will adopt the branch (keeping your authorship) so it can land for 1.7.0.

brandonrc:
Thanks for this fix, @garrettKelleyCyOne — the approach (consuming the backend's `?group_by=docker_tag` rollup instead of re-deriving tags from one page of the flat list) verified clean against the backend contract.

Since we didn't hear back on the adoption offer, we've adopted this PR for the 1.7.0 cut: your commit is preserved as-is (authorship intact), rebased onto current `main` with conflicts against #700 (quarantine badge contract) and #709 (folder-tree view) resolved so both intents are kept. Follow-up PR: #714. Leaving this one open for reference.

brandonrc:
Superseded by #714 (merged) — your commit landed with authorship preserved, plus coverage tests for the grouped-view callbacks. Thanks @garrettKelleyCyOne!

Issue #253, opened 2026-04-12
Show artifact quarantine status in artifact list and detail views


brandonrc:
Backend PR artifact-keeper/artifact-keeper#704 added a quarantine period for newly uploaded artifacts. The frontend should:

1. Show quarantine status badge on artifacts (quarantined/released/rejected)
2. Show remaining quarantine time for quarantined artifacts
3. Add admin actions: manual release and manual reject buttons
4. Prevent download button for quarantined artifacts (show 'under review' message)

API endpoint: GET `/api/v1/quarantine/artifacts/:id/quarantine`
Admin endpoints: POST `.../release`, POST `.../reject`
Config available via GET `/api/v1/system/config` (quarantine_enabled field)

PR #654, opened 2026-07-28
feat(quarantine): release and reject a held artifact from the web UI

### LINKED ISSUE #650: No UI to release or reject a quarantined artifact (#253 closed with admin actions unimplemented)
## Summary

There is no way to release or reject a quarantined artifact from the web UI. Every quarantine operation is API-only today.

`#253` covered this and was closed on 2026-04-19, but two of its four acceptance criteria were never implemented.

## What `#253` asked for, and what shipped

| Item from `#253` | Status |
|---|---|
| 1. Quarantine status badge on artifacts | Partial. `QuarantineBadge` exists and renders in search results only |
| 2. Remaining quarantine time | `formatQuarantineExpiry` exists in `src/lib/quarantine.ts` |
| 3. Admin actions: manual release and reject buttons | **Not implemented** |
| 4. Prevent download for quarantined artifacts, show "under review" | **Not implemented** |

Item 1 additionally does not work on the repositories page. `QuarantineBanner` is rendered in the artifact detail dialog behind `isActivelyQuarantined(selectedArtifact)`, but `selectedArtifact` comes straight from the artifacts list row, and `adaptArtifact()` leaves the quarantine fields undefined by design:

```ts
// Local Artifact extends ArtifactResponse with quarantine fields the SDK
// doesn't model yet — leave those undefined and let callers fetch detail
// endpoints if they need quarantine state.
```

No caller fetches the detail endpoint, so the guard is always false and the banner never mounts. The backend side of this is tracked in artifact-keeper/artifact-keeper#2940, where the listing returns `is_quarantined: null` for artifacts that are actively blocked.

## Current state of the surface

Grepping `src/lib/api/` finds no call to either admin endpoint. The backend has had them all along:

```
GET  /api/v1/quarantine/{artifact_id}          status
POST /api/v1/quarantine/{artifact_id}          admin hold
POST /api/v1/quarantine/{artifact_id}/release  admin release
POST /api/v1/quarantine/{artifact_id}/reject   admin reject
```

`release` and `reject` are admin-gated with `require_admin()`.

## Proposal

1. Wire release and reject into the artifact detail dialog, admin only, matching the pattern the age-gate review queue already uses for its approve and reject actions.
2. Make `QuarantineBanner` actually reachable, either by fetching quarantine state for the selected artifact or by consuming the fields once the listing populates them.
3. Disable or replace the download control for a quarantined artifact and explain why, rather than letting the user click through to a 409.
4. Consider a queue view listing everything currently quarantined. The backend has no list endpoint today (all four routes are keyed by a single `artifact_id`), so this needs a backend counterpart first.

Note `reject` accepts a reason in its request body; `release` does not. Worth confirming whether release should capture one too, since it is the action that unblocks a package a policy flagged.

## Notes

Distinct from `#549`, which covers the publish-age gate review queue. As that issue already notes, the `quarantine-badge` component belongs to the older quarantine feature, not the age gate.

Reopening `#253` instead of filing fresh would also be reasonable.

brandonrc:
> **Better sequenced after artifact-keeper/artifact-keeper#2958, but not blocked by it.** This PR reads per-artifact quarantine fields that the backend's `main` does not return yet. It degrades rather than breaks: absent fields resolve to `unknown` rather than `clear`, and the artifact detail dialog falls back to `GET /api/v1/quarantine/{id}`, which has been on `main` all along. Against an older backend the listing simply carries no quarantine badges. Merging #2958 first is preferable so the listing is populated, but it is not a hard ordering requirement.

## Summary

Every quarantine operation was API-only. #253 shipped a `QuarantineBadge` and a `QuarantineBanner` but never built the admin actions, and the banner was unreachable: it was gated on `isActivelyQuarantined(selectedArtifact)`, the artifact came straight from a listing row, and `adaptArtifact` left the quarantine fields undefined by design. Nothing ever populated them, so the guard was always false.

The backend now returns per-artifact quarantine state on the repository artifacts listing (artifact-keeper/artifact-keeper#2940): `is_blocked`, `quarantine_status`, `quarantine_until` and `quarantine_reason` on every row. All four are omitted rather than nulled on surfaces that do not load quarantine state, so absent means "the server did not look" while `false` means "it looked and this is downloadable".

Collapsing those two into one falsy check is the bug that made the banner dead code, so the distinction is now explicit: `quarantineKnowledge()` answers `unknown` / `clear` / `blocked`, and callers that would otherwise assert an artifact is fine go through it. An artifact opened through the by-path detail endpoint, which carries none of the four keys, resolves its verdict with `GET /api/v1/quarantine/{id}` instead of being assumed clean.

The field is `is_blocked`, not the `is_quarantined` the web read. The backend never had a field by that name.

Admin actions live in the artifact detail dialog, structured like the age-gate review queue's approve and reject so the two review surfaces behave the same way: a confirmation dialog per decision, an optional reason on reject, and the buttons gated on `is_admin` in addition to the backend's `require_admin`. A rejection is terminal, so neither action is offered once an artifact has been rejected, where the request could only come back 409. The reason is redacted for callers who cannot access the repository, so the banner always states the hold and treats the reason as an extra line.

The three admin endpoints are not in the generated SDK, so they go through the shared `apiFetch` wrapper, the same escape hatch `repositoriesApi.setReleaseTarget` uses, rather than blocking on an SDK regen.

## Test Checklist
- [ ] Manually tested locally

`npx vitest run`: 155 files, 2891 tests, all passing. `npm run lint`: 0 errors.

Not manually tested: no backend build carrying artifact-keeper#2940 exists yet, so the new fields cannot be exercised end to end until that merges and an image is published. This is the main outstanding risk on this PR and the reason for the merge-order note above.

## UI Changes
- [ ] Dark mode verified
- [ ] N/A, no UI changes

The four boxes above are unchecked deliberately rather than being claimed. The dialog reuses existing shadcn primitives that already carry the project's responsive, dark-mode and focus behaviour, but none of it has been verified in a browser against a backend that returns these fields.

Closes #650


brandonrc (commented):
Maintainer review — posted after merge (see process note at the end).

Thanks, @brandonrc — this is a careful piece of work: the admin-action flow mirrors the age-gate review queue, the `unknown`/`clear`/`blocked` tri-state in `src/lib/quarantine.ts` is the right fix for the dead-banner regression, the reason-redaction handling matches the backend's disclosure rules, and the test suite (API contract tests, dialog behavior, download gating) is genuinely thorough. I verified every endpoint against the backend handlers in `backend/src/api/handlers/quarantine.rs`:

- `GET /api/v1/quarantine/{id}`, `POST .../release`, `POST .../reject`, `POST .../quarantine` — paths, methods, admin gating, and the 409 transition rules all match (`src/lib/api/quarantine.ts`).
- Response shapes match: `QuarantineStatusResponse` (`quarantine_reason` omitted via `skip_serializing_if` when withheld) and `QuarantineActionResponse` (`artifact_id` / `new_status` / `message`).
- `reject`'s `{reason: null}` body is valid for the backend's required `Json<RejectRequest>` extractor, and `release`'s empty `{}` is harmless (no body extractor).

One correctness problem needs a follow-up PR, because the backend contract this was written against is not the one that shipped:

**The listing payload on backend `main` carries no `is_blocked`.** This PR targets artifact-keeper/artifact-keeper#2958 — which is still open. What actually merged (2026-07-28, three days before this PR) is artifact-keeper/artifact-keeper#2966, with a different shape:

- `ArtifactResponse.quarantine_status` is **always serialized** — `"quarantined"`, `"rejected"`, a scan-lifecycle state (`"clean"` / `"flagged"` / `"unscanned"` / `"released"`), or the explicit `"not_quarantined"` default.
- `quarantine_until` is present only for timed holds.
- There is **no `is_blocked` and no `quarantine_reason` on the listing** — the reason stays behind the authenticated, visibility-checked `GET /api/v1/quarantine/{id}` by design.
- The by-id **and by-path** artifact responses *do* carry `quarantine_status` / `quarantine_until`, so the premise in `repo-detail-content.tsx` that the by-path endpoint "omits all four keys" no longer holds either.

Consequences against the shipped backend:

1. `src/lib/api/artifacts.ts:76` copies a field the server never sends, so `quarantineKnowledge()` (`src/lib/quarantine.ts:62`) returns `unknown` for every listing row. `docker-tag-list.tsx:224` therefore never renders `QuarantineBadge` and suppresses "OK" for *all* tags — issue #650 item 1 is still dead on the listing, just silently.
2. `repo-detail-content.tsx:274` — `isQuarantineStateKnown(selectedArtifact)` is always false on listing rows, so the `GET /quarantine/{id}` fallback fires on every dialog open. Release/reject still function through that path (which is why the tests pass — they mock the unshipped shape), but the listing-driven primary path this PR was built around is inert, and every dialog open costs an extra request.
3. The tests in `__tests__/repo-detail-quarantine.test.tsx` and `lib/api/__tests__/quarantine.test.ts` pin the #2958 shape, so they will keep passing while the UI reads nothing the backend sends.

Requested follow-up (happy to review quickly): derive the verdict from the always-present `quarantine_status` — `"quarantined"` (respecting `quarantine_until` expiry, same semantics as `check_download_allowed` in `quarantine_service.rs`) and `"rejected"` are blocked; `"not_quarantined"` and scan-lifecycle states are clear — and fetch the reason lazily via `GET /api/v1/quarantine/{id}` only for blocked artifacts, keeping the redacted-reason fallback you already built.

Smaller follow-ups, same PR or separate:

- `src/lib/api/quarantine.ts` doesn't validate `apiFetch` responses at the trust boundary — the convention for non-SDK modules is a zod schema or explicit parse (see `src/lib/api/audit.ts`, `downloads.ts`, `parseBlastRadius` in `blast-radius.ts`).
- The comment in `reasonBody()` is slightly off: `reason` is an `Option<String>`, so an absent key deserializes fine — it's the JSON *body* that is required. Sending `{reason: null}` is correct; the rationale isn't.
- (The default export was already normalized to `export const quarantineApi` by #694, and `unwrap()` / `useRepositories` / `ListTruncationNotice` all landed after this PR — nothing to do there.)

Process: this was merged by its author with zero reviews while a `web-ui` team review was still requested — and three days after the backend shipped a different contract than the one in the PR description. Please leave the merge to a maintainer next time; a review here would have caught the mismatch before it landed on `main`.

— Maintainer review (artifact-keeper-web)

PR #655, opened 2026-07-28
feat(age-gate): let admins change a review decision after the fact

### LINKED ISSUE #651: Age-gate review queue: cannot change a decision once approved or rejected, plus follow-up rework
## Summary

The age-gate review queue page needs a pass beyond the initial implementation. The immediate gap: once a review is approved or rejected there is no way to change it from the UI, and no indication that the decision is permanent.

## Blocked on the backend

A status control cannot be built yet. Approved and rejected are terminal states in the API, and both mutating endpoints reject any non-pending review:

```
POST /api/v1/admin/age-gate/reviews/{id}/approve
  HTTP 400  {"code":"VALIDATION_ERROR","message":"Review is already approved"}

POST /api/v1/admin/age-gate/reviews/{id}/reject
  HTTP 400  {"code":"VALIDATION_ERROR","message":"Review is already approved"}
```

There is no PATCH, PUT, or DELETE on the review resource. A dropdown has nothing to call until the backend adds a reopen or set-status endpoint.

Blocked by artifact-keeper/artifact-keeper#2939.

## Wanted, once unblocked

- A status control on each row, so an admin can move a review back to pending or flip a decision, with the reason captured.
- Clear signalling of which transitions are available. If a state stays terminal by design, the UI should say so rather than offering an action that 400s.

## Independent of the backend

These can proceed now:

- Surface the reason text recorded with an approve or reject decision. The API accepts one and it is not currently shown.
- Show who decided and when, since the backend already writes `AgeGateApproved` to the audit log with the user id.
- Filtering by status. `list_reviews` already accepts a `statuses` parameter that the UI does not expose.
- Empty, loading, and error states for the queue.
- Confirmation before approving, since approval is currently irreversible.

## Notes

Extends `#549`, which covers the original age-gate console work. Filing separately rather than reopening because `#549` scopes the initial build and this is follow-up on the shipped page. Happy to fold it into `#549` instead if that is preferred.

brandonrc:
> **Do not merge before artifact-keeper/artifact-keeper#2957.** The status control calls a reopen endpoint that does not exist on the backend's `main` yet. Merged on its own, every status change returns 404.

## Summary

Once a review was approved or rejected the queue offered no way back. The row's only controls were a one-way approve and a one-way reject, both of which 400 on a decided review, and nothing on the page said the decision was permanent. An admin who approved the wrong package under time pressure had no recovery path short of editing `age_gate_reviews` by hand.

The status is now a dropdown carrying every state, backed by the reopen endpoint artifact-keeper/artifact-keeper#2939 adds. Moving a decided review to `pending` reopens it, which re-applies the gate and withholds the version again. Moving it straight to the other decision is reopen-then-decide: approve and reject only accept a pending review, so one operator action is two calls.

That composition has a real halfway state, and it is the case most able to mislead. If the reopen lands and the decision fails, the review is neither where it started nor where it was headed: it is `pending`, and the package is withheld. `changeReviewStatus` raises `AgeGatePartialTransitionError` for exactly that path, and the page refetches and names the status the review is actually in, rather than reporting a failure that implies the original decision still stands.

Reopen requires a non-blank reason, validated before the call so a blank one does not cost a round trip to learn nothing. Approve and reject keep an optional reason. Approving now confirms against the package being released, and the dialog says which transitions run in two steps and where they stop if the second fails.

The rest is metadata the page was already receiving and discarding. The recorded reason, the deciding admin and the decision date now have a column; the response carries only a user id, so names come from the admin user list and fall back to a shortened id when the lookup misses. Status filtering moved to checkboxes that send the backend's comma-separated multi-status filter, with its own loading, empty, and error states.

`reopen` is not in the generated SDK, so it goes through `apiFetch` like routing-rules and audit do, rather than blocking on a regen.

## Test Checklist
- [ ] Manually tested locally

`npx vitest run`: 153 files, 2869 tests, all passing. `npm run lint`: 0 errors. The reopen-then-decide partial-failure path has its own test, since that is the one that can mislead an operator.

Not manually tested: no backend build carrying artifact-keeper#2939 exists yet, so the reopen path cannot be exercised end to end until that merges and an image is published. This is the main outstanding risk on this PR and the reason for the merge-order note above.

## UI Changes
- [ ] Dark mode verified
- [ ] N/A, no UI changes

Unchecked deliberately rather than claimed. The controls reuse existing shadcn primitives that already carry the project's responsive, dark-mode and focus behaviour, but none of it has been verified in a browser against a backend that has the reopen endpoint.

Closes #651


brandonrc (commented):
Post-merge review — this landed before anyone got eyes on it, so filing findings here for follow-up rather than blocking. Overall the tests are genuinely good (the partial-transition case, the capability-latch cases, and the blank-reason guard all earn their keep), the `ApiError` addition is cleanly backward compatible, and the capability heuristic is sound (I verified an unrouted path returns the framework's empty-body 404 while the handler's unknown-id 404 carries the `{"code":"NOT_FOUND",...}` envelope). A few things need a follow-up PR, one of them functional:

**1. The reopen-then-decide composition contradicts the backend's design, and it opens a real exposure window.** The premise in the description and in `changeReviewStatus` (`src/lib/api/age-gate.ts:285`) — "approve and reject only accept a pending review, so one operator action is two calls" — is not true of the backend that actually shipped. artifact-keeper#2968 deliberately relaxed this: `require_distinct_status` (`backend/src/services/age_gate_service.rs:190`) refuses only same-status no-ops, so `approve` accepts a `rejected` review and `reject` accepts an `approved` one, in a single call. Consequences:

- The entire `AgeGatePartialTransitionError` path exists only because of the unnecessary two-call split. A decided→decided change as one call has no halfway state at all.
- Worse, the halfway state is not neutral. Reopen moves the review to `pending`, and the gate only withholds releases younger than `min_age_days` — the backend service doc says reopening re-blocks "until it is decided again (or ages out)". For a version that has since aged past the threshold, the window between the reopen call and the decide call makes it downloadable. A direct `approved → rejected` call never exposes it. For an age-gate feature that window is exactly the wrong direction to err.

Suggest: decided→decided goes straight to `approveReview`/`rejectReview` in one call; keep `reopenReview` only for the →`pending` case. That deletes the partial-transition machinery, the "two steps" dialog copy, and the exposure window in one move.

**2. The `reviewedBy` docblock is wrong, and the Decision column will mislead on reopened rows.** `src/lib/api/age-gate.ts:47-49` says the backend clears the reviewer on reopen so a reopened row doesn't read as already handled. It doesn't: `reopen` sets `reviewed_by` to the reopening admin and `reviewed_at = NOW()` (`age_gate_service.rs:668`). So a reopened, now-pending review renders in the Decision column (page.tsx:329) as "decided by <whoever reopened> just now — '<reopen reason>'" instead of "Not yet decided". Either filter the Decision cell on `status !== 'pending'` or fix the comment and accept the display — but right now the code says one thing and does another.

**3. The blank-reason justification is also wrong, mildly.** The description says the backend 400s on a blank reopen reason. There is no server-side validation — `ReviewActionRequest.reason` is `Option<String>` and neither the handler nor `svc.reopen` checks for blank. The client-side guard is good UX regardless, but if a non-blank reason is an audit requirement it has to live in the backend; worth a note on the backend repo.

**4. Silent truncation at 20 rows.** `list_reviews` defaults to `per_page = 20` (max 100) and the page (`page.tsx:131`) discards the pagination envelope entirely. The multi-status checkbox filter makes queues over 20 rows much more likely than the old single-status select did. The house pattern for this is `ListTruncationNotice` — please pass `perPage: 100` and render the notice from `pagination.total` (or paginate for real). Relatedly, `adminApi.listUsers()` (`src/lib/api/admin.ts:108`) only fetches the first page of users, so reviewers outside it fall back to the shortened id — acceptable degradation, but worth a comment in the code so nobody reads it as a bug later.

**5. Stale dependency note.** The merge-order warning pointed at artifact-keeper#2957, which is still open and appears to be a superseded duplicate — the endpoint actually shipped via artifact-keeper#2968, which is merged. So the `ReopenUnsupportedError` latch will simply never fire against a current backend. Harmless; just flagging that the "do not merge before" concern was moot.

**On process:** this was self-merged with zero reviews and one red check. The red shard (E2E Interactions 3/3) was an infra flake — a docker.io timeout pulling the meilisearch image, not a test failure — but that's exactly the call a second pair of eyes should be making. Please hold for review even when you can green-button yourself.

**On #452:** this PR supersedes it only partially. The review-queue half of #452 (the `/age-gate` page and `lib/api/age-gate.ts`) is now on main via #635 + this PR. But #452 also carried the per-repo `AgeGateSettings` panel and the `age_gate_*` webhook-event wiring, neither of which exists on main today, and it's built against the backend-#1403-era API so a rebase won't save it. My recommendation: #452 should be closed as superseded-in-part, with the settings panel and webhook events re-scoped as fresh PRs against current main. Leaving that note here rather than on #452 itself.

— maintainer review

PR #657, opened 2026-07-30
Fix package age policy state not persisting across navigation (#665)

### LINKED ISSUE #656: Package Age Policy not working
### Description

When enabling package age policy and/or changing cooldown period in repository settings it's respond as policy saved. After change page and returning into repository settings it's shows as disabled and cooldown period sets as default 3 days

### Environment Information

- Browser: Chrome
- Browser Version: 144.0.7559.133
- Operating System: Windows 10
- Screen Resolution: 2560x1080

### Web UI Version

1.6.0

### Backend Version

1.6.0

### Expected Behavior

Package age policy enabled as cooldown period sets

### Observed Behavior

After change page and returning into repository settings it's shows as disabled and cooldown period sets as default 3 days

### Steps to Reproduce

1. Navigate to repositories. Select repository. 
2. Click on settings
3. Scroll down to package age policy
4. Click on enable age policy. 
5. Click on cooldown period and set another value - not 3 days
6. Click on save age policy
7. Select another repository
8. Return to change repository settings

### Browser Console Logs

```shell

```

### Screenshots

_No response_

### Additional Context

_No response_

### Pre-submission Checklist

dvodop:
The age policy form in repository settings was initialised with hardcoded defaults (disabled, 3 days) and never read the persisted server state, so after saving and navigating away the form would reset as if no change had been made.

Add `quarantine_enabled` and `quarantine_duration_minutes` to the Repository type and seed the age policy form state from the repository object using the same override pattern as the general settings fields. This ensures the form reflects the true server state after mount and after a save-then-refetch cycle.

Closes #656


brandonrc (approved):
Verified against the backend and the current conventions — this is correct and well-tested. Approving; a few notes below, none blocking.

**Issue reference:** the title, commit headline, and the new code comment in `repo-settings-tab.tsx` ("regression #665") point at #665, but #665 was the maven grouped-browser issue (already fixed by #696). The PR body correctly says `Closes #656` ("Package Age Policy not working"), which is the issue this actually fixes — the auto-close will do the right thing. Please fix the comment reference to #656 when you next touch the branch; Ill retitle the PR on merge.

**Contract check (backend, verified):** `RepositoryResponse` carries `quarantine_enabled: Option<bool>` / `quarantine_duration_minutes: Option<i64>` (`backend/src/api/handlers/repositories.rs:819-823`), populated only by the detail and update handlers via `with_quarantine_settings` (lines 2800, 3650) and omitted from the JSON entirely when unset. The settings tab gets its `repository` from the single-repo GET (`repo-detail-content.tsx:190-191`), not the list (which intentionally omits these fields), and the `?? undefined` handling in `adaptRepository` matches the omission behavior — so the optional fields on the `Repository` type are the right shape. PATCH validation (non-negative duration) matches what the form sends.

**Convention check:** the `ageDefaults`/`ageOverrides` split is exactly the pattern the general settings, quota, debian, and npm sections of this same file already use, and resetting overrides + invalidating `["repository", key]` on success re-seeds cleanly from the refetched query data. This is the right approach here (TanStack Query as source of truth, local state only for unsaved edits) — preferable to a render-time sync.

**Overlap with #661:** heads-up that #661 (for #658) contains an identical age-policy half — same `minutesToAge` helper, same `adaptRepository` and `Repository` type additions — plus its release-target work. Since this PR is the focused fix for #656, was opened first, and is mergeable while #661 is already dirty against main, I plan to land this one first; #661 will then need to drop its duplicated age-policy hunk on rebase (conflicts will be in `repo-settings-tab.tsx`, `repo-settings-tab.test.tsx`, `lib/api/repositories.ts`, `lib/api/__tests__/repositories.test.ts`, and `types/index.ts`).

Non-blocking nits:
- `repo-settings-tab.tsx` — `ageDirty` is derived from key presence in `ageOverrides`, so toggling a value and then back to the seeded value still counts as dirty and Save re-writes identical values. Harmless; the general settings section compares values instead, if you ever want to tighten it.
- `minutesToAge` rounds sub-day remainders (e.g. 150 min → "3 hours"); if someone set an odd duration via the API and then toggles the switch without touching the field, a save would silently write 180. Edge case, fine to leave.
- `e2e/suites/interactions/repositories/age-policy.spec.ts:50-59` — the reset PATCH swallows failures with `.catch(() => {})` and never asserts `ok()`, so a failed reset leaves the UI test running against unknown state. It matches the pre-existing reset pattern in this file, so no action needed, but a status assert would make flakes easier to diagnose.

Coverage is solid: `minutesToAge` unit tests, component tests for seeding/pristine/dirty transitions, adapter pass-through tests, and the e2e hardening. CI is all green.

PR #660, opened 2026-07-31
feat(lifecycle): select repository scope in UI

### LINKED ISSUE #659: Lifecycle policy form cannot select the repository required by scoped policies
## Description

The Lifecycle Policy form does not expose a repository selector for policy types
that require a repository scope.

`max_versions` and `size_quota_bytes` require `repository_id` as a top-level
API field. The UI only exposes the policy JSON configuration, so an
administrator cannot create these policies from the web interface.

## Environment Information

- Browser: Chrome
- Operating System: macOS
- Connection type: direct
- Web UI Version: 1.6.0

## Expected Behavior

When an administrator selects **Max Versions** or **Size Quota**, the form should:

- show a searchable repository selector;
- require a repository before enabling Create; and
- submit its UUID as `repository_id`, separately from the JSON configuration.

Policies that are valid globally must remain creatable without a repository.

## Observed Behavior

The form has no repository selector. Creating a scoped policy from the UI fails
because the required `repository_id` is not sent.

## Steps to Reproduce

1. Open **Admin → Lifecycle Policies**.
2. Select **New Policy**.
3. Choose **Max Versions** or **Size Quota**.
4. Enter a name and valid JSON configuration.
5. Observe that no repository can be selected and creation is rejected.

## Pre-submission Checklist

- [x] I searched existing issues and pull requests for duplicates.
- [x] I redacted sensitive information.
- [x] I tested the behavior on the current web UI.

ivolnistov:
## Summary

Closes #659

Adds a searchable repository selector to the Lifecycle Policy form for
repository-scoped policy types:

- loads repositories only when creating **Max Versions** or **Size Quota**;
- requires a selection before the policy can be created;
- sends the selected UUID as the top-level `repository_id` API field; and
- keeps globally applicable lifecycle policies repository-independent.

The selector filters by repository key, name, format, and type, and displays
the repository key together with its format and type.

## Validation

- `npm run lint` — completed with no errors (existing repository warnings remain)
- Targeted Vitest lifecycle form tests — 3 passed
- Playwright coverage added for request payload, keyboard selection, and mobile layout
- Manually verified on a production deployment: desktop and 390px mobile layout,
  keyboard search/selection, dark mode, and accessibility roles/names

## Test Checklist

- [x] Manually tested on a production deployment
- [x] Targeted tests pass without regressions

## UI Changes

- [x] Dark mode verified
- [ ] N/A - no UI changes

### DIFF
--- a/e2e/suites/interactions/operations/lifecycle.spec.ts
+++ b/e2e/suites/interactions/operations/lifecycle.spec.ts
@@ -151,17 +151,18 @@ test.describe('Lifecycle Page', () => {
     const repositoryPicker = dialog.getByRole('combobox', { name: 'Repository' });
     await expect(repositoryPicker).toHaveAccessibleName('Repository');
     await expect(repositoryPicker).toBeEnabled();
-    await repositoryPicker.focus();
-    await page.keyboard.press('Enter');
-    const repositorySearch = page.getByLabel('Search repositories');
-    await expect(repositorySearch).toHaveAccessibleName('Search repositories');
+    await repositoryPicker.click();
+    const repositorySearch = page.getByRole('combobox', {
+      name: 'Search repositories',
+    });
+    await expect(repositorySearch).toBeVisible();
     await repositorySearch.fill(repository!.key);
-    const repositoryOption = page
-      .locator('[data-slot="command-item"]')
-      .filter({ hasText: repository!.key });
+    const repositoryOption = page.getByRole('option', {
+      name: repository!.key,
+      exact: false,
+    });
     await expect(repositoryOption).toBeVisible();
-    await page.keyboard.press('ArrowDown');
-    await page.keyboard.press('Enter');
+    await repositoryOption.click();
 
     await expect(repositoryPicker).toContainText(repository!.key);
     await dialog.getByRole('button', { name: /^create$/i }).click();
--- a/e2e/suites/interactions/operations/lifecycle.spec.ts
+++ b/e2e/suites/interactions/operations/lifecycle.spec.ts
@@ -1,5 +1,34 @@
 import { test, expect } from '@playwright/test';
 
+type RepositoryListResponse = {
+  items: Array<{ id: string; key: string }>;
+};
+
+type LifecycleCreateRequest = {
+  repository_id?: string;
+  name?: string;
+  policy_type?: string;
+  config?: Record<string, unknown>;
+};
+
+function lifecyclePolicyStub(body: LifecycleCreateRequest) {
+  const timestamp = new Date().toISOString();
+  return {
+    id: '00000000-0000-4000-8000-000000000001',
+    repository_id: body.repository_id ?? null,
+    name: body.name ?? 'e2e lifecycle policy',
+    description: null,
+    enabled: true,
+    policy_type: body.policy_type ?? 'max_versions',
+    config: body.config ?? { keep: 1 },
+    priority: 0,
+    last_run_at: null,
+    last_run_items_removed: null,
+    created_at: timestamp,
+    updated_at: timestamp,
+  };
+}
+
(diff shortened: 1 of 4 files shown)


brandonrc (changes requested):
Thanks for tackling #659 — the UX here is well thought out, and the test coverage (payload-asserting E2E, keyboard selection, mobile viewport, error/empty states in the unit tests) is genuinely above the bar we usually see. I verified the API contract against the backend and the payload shape is exactly right: `repository_id` is a top-level field on `POST /api/v1/admin/lifecycle`, and the backend rejects `max_versions` / `size_quota_bytes` policies created without one, so `POLICY_TYPES_REQUIRING_REPOSITORY_ID` in `src/types/lifecycle.ts` matches server-side validation. Gating the repositories query behind `createOpen && requiresRepositoryId` is a nice touch too.

A few things need addressing before this can merge, so I'm requesting changes — mostly collision with work that landed on main after this branch was cut:

**1. Rebase required — `lifecycleApi` is now a named export (blocking).**
This branch is ~19 commits behind main. #694 landed in the meantime and standardized `src/lib/api/lifecycle.ts` on a named export (`export const lifecycleApi = {...}`); main's lifecycle page now does `import { lifecycleApi } from "@/lib/api/lifecycle"`. Your `import lifecycleApi from "@/lib/api/lifecycle"` in `src/app/(app)/(admin)/lifecycle/page.tsx:20` sits on the exact line that changed, so this will conflict on merge and fail to build as written. Please rebase onto current main, switch to the named import, and update the corresponding mock in `page.test.tsx` (`vi.mock("@/lib/api/lifecycle", () => ({ default: lifecycleApi }))` → `({ lifecycleApi })`).

**2. Use the shared `useRepositories()` hook instead of a hand-rolled query.**
`src/hooks/use-repositories.ts` (#669) is now the canonical way to fetch repository lists — it centralizes the cache key via `repositoryListKey(params)` so every consumer shares one cache entry and prefix invalidation works uniformly. The ad-hoc `useQuery({ queryKey: ["repositories", "lifecycle-policy-selector"], queryFn: () => repositoriesApi.list({ per_page: 1000 }) })` in `page.tsx` (~line 133) should become:

```ts
const { data: repositoriesPage, isLoading: isLoadingRepositories, isError: repositoriesError } =
  useRepositories(
    { per_page: 1000 },
    { enabled: !!user?.is_admin && createOpen && requiresRepositoryId },
  );
```

Your existing unit-test mock dispatches on `queryKey[0] === "repositories"`, which still matches the canonical key, so the tests should only need minimal adjustment.

**3. The selector silently truncates beyond 100 repositories.**
The backend clamps `per_page` to a maximum of 100, so `per_page: 1000` returns at most 100 repositories — on installations with more, repos past the first page simply won't appear in the picker, with no indication to the user (and the Create button stays disabled if the intended repo isn't listed). Please surface this with the `ListTruncationNotice` component when `repositoriesPage.total > repositories.length` — see `curation-rules-manager.tsx` or `virtual-members-panel.tsx` for the established pattern.

**CI note:** shard 3 is red, but the failure is `maven-grouped-browser.spec.ts` (a spec this PR doesn't touch), and earlier pushes failed on a different unrelated spec, so this looks like E2E flakiness rather than something you introduced — your new lifecycle E2E tests passed. Still, we can't merge on red CI; a rebase + re-run should sort it out. If `maven-grouped-browser` fails again after the rebase, flag it and we'll investigate it as a flake separately.

One small non-blocking observation: the edit flow doesn't expose repository scope (policies keep whatever scope they were created with) — that's fine for this PR since the backend doesn't allow changing it on update either, but worth a follow-up issue if we ever want that in the UI.

Happy to re-review quickly once the rebase and the three points above are in.

— Maintainer review

PR #662, opened 2026-07-31
fix(security): choose a repository in the policy form instead of a raw UUID (#489)

### LINKED ISSUE #489: Security Policies form asks for a Repository UUID but the UI nowhere exposes it — needs a repo Select
## Problem

The Security Policies create/edit form (`/security/policies`) asks the user to type a repository UUID into a free-text input:

```tsx
// src/app/(app)/(admin)/security/policies/page.tsx:172-190
<Label htmlFor="policy-repo-id">Repository ID (optional)</Label>
<Input
  id="policy-repo-id"
  placeholder="Leave blank for a global policy"
  value={form.repository_id}
  onChange={(e) => setForm((f) => ({ ...f, repository_id: e.target.value }))}
/>
<p className="text-xs text-muted-foreground">
  Scope this policy to a specific repository, or leave blank for global
  enforcement.
</p>
```

**The web UI exposes no place to find that UUID.** I audited every repository-related page:

| Surface | Repo UUID visible? |
|---|---|
| Repositories list (`/repositories`) | No — shows `name`, `key`, `format` only |
| Repo detail URL | No — route uses the `key` (`/repositories/ci-maven-hosted`) |
| Repo detail tabs (Settings / Security / Packages / Notifications) | No — `repository.id` is consumed internally for API calls (`repo-detail-content.tsx:265,889`, `repo-settings-tab.tsx:284`) but never rendered |
| "Copy ID" button anywhere | No — entire repo has one `navigator.clipboard` call and it's in `package-metadata-viewer.tsx` for metadata, not repo IDs |
| Policies list page | Truncated only — `Repo: 00827aab...` (first 8 chars, not selectable, not copyable) — `policies/page.tsx:284-295` |

So the form requires data that the UI hides from the user. The only way to obtain the UUID today is:

1. Open browser DevTools → Network tab → inspect a request body for `repository_id`
2. Hit the API: `curl /api/v1/repositories | jq '.items[].id'`
3. Query the DB

None of those are reasonable for a UI-driven workflow.

## Proposed fix

Replace the free-text `<Input>` with a `<Select>` that lists repositories by `key`+`name`, with a "Global" entry corresponding to `null` (the current "leave blank" semantic).

The exact pattern already exists in this codebase for the release-target picker (`src/app/(app)/repositories/_components/release-target-settings.tsx:155-167`):

```tsx
<Select value={selected} onValueChange={handleSelect}>
  <SelectTrigger><SelectValue placeholder="..." /></SelectTrigger>
  <SelectContent>
    <SelectItem value={NONE_VALUE}>Global (all repositories)</SelectItem>
    {(repoList?.items ?? []).map((repo) => (
      <SelectItem key={repo.id} value={repo.id}>
        {repo.name} ({repo.key})
      </SelectItem>
    ))}
  </SelectContent>
</Select>
```

Differences for the policy form:
- The "None" entry is **kept** (it maps to `null` → global policy) — not "unlink" semantics
- `value` is `repo.id` (UUID), not `repo.key`, because that's what the backend expects (`CreatePolicyRequest.repository_id: Option<Uuid>` in `backend/src/api/handlers/security.rs:388-399`)
- Display label is `{repo.name} ({repo.key})` so the user sees human text but the form value is still the UUID

Use `repositoriesApi.list({ per_page: 200 })` (same call pattern release-target-settings already uses). No backend change required.

## Bonus — make the policies list readable

`policies/page.tsx:284-295` currently shows `Repo: 00827aab...` for non-global rows. Once the API is being fetched anyway (for the picker), the list cell can resolve UUID → name and show `Repo: CI Maven Hosted` instead. The first-8-char prefix is unrecognizable to anyone who isn't already looking the UUID up by hand.

## Reproduction

1. Go to `/security/policies` as an admin
2. Click "New Policy"
3. See the "Repository ID (optional)" text field
4. Realize you have no way to find a repo's UUID from the UI
5. Either give up and create only Global policies, or open DevTools / curl to extract one

## Acceptance criteria

- [ ] Create/edit Policy form replaces the Repository ID `<Input>` with a `<Select>` populated from `GET /api/v1/repositories`
- [ ] Select includes a "Global (all repositories)" option that maps to `null`
- [ ] Submit payload still sends `repository_id: <uuid>` or `repository_id: null` (matches `CreatePolicyRequest`)
- [ ] Policies list cell renders the resolved repo name + key for scoped policies, not a truncated UUID
- [ ] Form continues to support raw-UUID entry for power users? **Optional** — could be implemented as a "Paste UUID" toggle below the Select. Not blocking.

## Workaround for current operators

```bash
# Get a UUID for a known repository key
TOK=$(curl -sk -X POST https://<host>/api/v1/auth/login \
  -H 'Content-Type: application/json' \
  -d '{"username":"admin","password":"..."}' | jq -r .access_token)

curl -sk https://<host>/api/v1/repositories \
  -H "Authorization: Bearer $TOK" | \
  jq -r '.items[] | "\(.key)\t\(.id)"'
```

Then paste the second column into the form.

## Related

- Surfaced during AK PoC verification on `kychen-1` cluster, 2026-06-15, while building Security Policies on top of the Replication / DT integration work
- Same UX pattern (form asks for a backend-only identifier the UI doesn't expose) exists in a few other places — feel free to fold them into a broader "expose backend IDs in the UI where forms need them" tracking issue if there's appetite, or keep this scoped narrow.
- Companion DT issues: artifact-keeper/artifact-keeper#1904, artifact-keeper/artifact-keeper-iac#168, artifact-keeper/artifact-keeper-iac#169
- Prior UI gaps issue from this PoC: #488

nicola-preda:
## Summary
Fixes #489.

The Security Policies form asked you to type a repository UUID into a free-text field, but the UI exposes that UUID nowhere (you had to dig it out of the API or DevTools), so in practice you could only create Global policies.

- Create form: the free-text "Repository ID" input is now a repository `<Select>` with a "Global (all repositories)" option plus each repo, sourced from `repositoriesApi.list`. Global submits `repository_id: null`; a chosen repo submits its UUID.
- Policies list: a scoped policy now shows the repo name/key instead of a truncated `Repo: 00827aab...`.
- Labels collapse to just the key when a repo's name equals its key, so it reads `anti-abuse-builds` rather than `anti-abuse-builds (anti-abuse-builds)`.

Scope stays create-only, matching the backend: `UpdatePolicyRequest` has no `repository_id`, so the edit form intentionally doesn't offer a repo picker.

## Test Checklist
- [ ] Manually tested locally

## UI Changes
- [ ] Dark mode verified


nicola-preda:
@brandonrc Small MR fixing the repository selection (asking the user to input UUID is not the way) in the policy form. E2E failure non related

brandonrc (approved):
Reviewed: implements #489; change is scoped to the matching components and the required checks (Build/Lint/Unit) are green.

brandonrc (commented):
Maintainer review — post-merge verification pass.

**Verdict: the approval was warranted.** This fixes #489 end to end and is correct against the backend contract. Since it has already merged, leaving this as a documented verification with two follow-ups.

**Contract verification (against the backend repo):**
- `CreatePolicyRequest.repository_id: Option<Uuid>` (`backend/src/api/handlers/security.rs:510`) — `null` correctly maps to a global policy, and the service resolves `repository_id IS NULL` as global scope in policy evaluation (`policy_service.rs:90`). The `GLOBAL_VALUE` sentinel mapping at `src/app/(app)/(admin)/security/policies/page.tsx` (Select `onValueChange`) is right.
- `UpdatePolicyRequest` indeed has no `repository_id` (`security.rs:536`), so keeping the repo picker create-only matches the backend. Good call documenting that in the PR body.
- The list badge falling back to `Repo: <uuid8>...` when the repo isn't in the loaded page is a sensible degradation.

**Tests:** `page.test.tsx` covers the three behaviors that matter (label resolution in the list, Select replaces free-text, UUID vs `null` submission) with the same native-select stub convention used elsewhere in this directory. Solid coverage for a unit-tier change.

**Conventions:** the hand-rolled `useQuery({ queryKey: ["repositories", "policy-scope"], ... })` predated the shared `useRepositories` hook from #692; #692 already migrated this page to `useRepositories({ per_page: 200 })`, so main is consistent — no action needed here.

**Follow-up 1 (real gap, not blocking):** the backend clamps `per_page` to 100 (`repositories.rs::clamp_per_page`, `.clamp(1, 100)`), so `per_page: 200` silently returns at most 100 repos. On instances with more than 100 repositories the picker and the badge-resolution map are both incomplete, with no UI signal. The established pattern is `ListTruncationNotice` (`src/components/common/list-truncation-notice.tsx`, used in `replication/page.tsx:485`). Please open a small follow-up to render it when `pagination.total > items.length` on this page.

**Follow-up 2 (process):** this was merged 2 seconds after approval while `E2E Interactions (shard 3/3)` was still running; it then failed on `maven-grouped-browser.spec.ts:121` (#443 grouped-mode pagination). That failure is unrelated to this change — the security-policies spec passed in the same shard, and shard 3 is green on main — but please wait for the full E2E matrix before merging in the future, or re-run the shard to confirm flakiness.

Thanks for the clean, well-scoped fix — the `repoLabel` name==key collapse is a nice touch.

— Maintainer review

PR #685, opened 2026-07-31
fix(permissions): support service account principals

### LINKED ISSUE #621: Permissions modal is missing service account as principal type
### Description

The current permissions modal does not allow selecting the Principal Type service account, making it impossible to assign or edit permissions of service accounts which is needed since backend 1.6.0

### Environment Information

Artifact Keeper Web 1.5.8

### Web UI Version

1.5.8

### Backend Version

_No response_

### Expected Behavior

Being able to select principal type Service Account so new permissions can be added to service account or existing permissions edited

### Observed Behavior

You can only select user or group blocking a user from assigning permissions to service account. Adding new permissions does work through the api, but it seems principal_name is not being populated in the permissions request

### Steps to Reproduce

Navigate to permissions
Select principal type user
try to assign any new permissions
submit fails with principal with id not found error due to lacking principal_type: service_account in the request

### Browser Console Logs

```shell

```

### Screenshots

Image

### Additional Context

_No response_

### Pre-submission Checklist

ivolnistov:
Fixes #621

## Summary

- Add **Service Account** as a permission principal type and submit its actual type and ID to the API.
- Replace the principal dropdown with a searchable, type-filtered picker; service accounts can be found by their stable `svc-*` username or display name.
- Resolve service-account labels in existing permission rows when the backend omits `principal_name`.

## Why

The backend already supports `service_account`, but the UI could only create user or group grants. This prevented administrators from granting a service-account token repository access through the UI.

## Test Checklist

- [x] Manually tested locally

## UI Changes

- [x] Dark mode verified
- [ ] N/A - no UI changes

### DIFF
--- a/e2e/suites/interactions/admin/permissions.spec.ts
+++ b/e2e/suites/interactions/admin/permissions.spec.ts
@@ -42,6 +42,32 @@ test.describe('Permissions Management', () => {
     await dialog.getByRole('button', { name: /cancel/i }).click();
   });
 
+  test('can select a service account as the permission principal', async ({ page }) => {
+    await page.getByRole('button', { name: /create permission/i }).first().click();
+    const dialog = page.getByRole('dialog', { name: 'Create Permission', exact: true });
+    await expect(dialog).toBeVisible({ timeout: 10000 });
+
+    const principalType = dialog.locator('[role="combobox"]').filter({ hasText: /^User$/ });
+    await expect(principalType).toHaveCount(1);
+    await principalType.click();
+    await page.getByRole('option', { name: 'Service Account', exact: true }).click();
+
+    const principalPicker = dialog.getByRole('combobox', { name: 'Principal', exact: true });
+    await expect(principalPicker).toBeVisible();
+    await principalPicker.click();
+
+    const principalSearch = page.getByRole('combobox', { name: 'Search principals' });
+    await principalSearch.fill('e2e-ci-bot');
+
+    const seededServiceAccount = page.getByRole('option', { name: 'Service account for E2E tests' });
+    await expect(seededServiceAccount).toBeVisible({ timeout: 10000 });
+    await seededServiceAccount.click();
+    await expect(principalPicker).toContainText('Service account for E2E tests');
+
+    await dialog.getByRole('button', { name: /cancel/i }).click();
+    await expect(dialog).not.toBeVisible({ timeout: 10000 });
+  });
+
   test('dialog has action checkboxes for Read, Write, Delete, Admin', async ({ page }) => {
     await page.getByRole('button', { name: /create permission/i }).first().click();
     const dialog = page.getByRole('dialog');
--- a/src/app/(app)/(admin)/permissions/page.test.tsx
+++ b/src/app/(app)/(admin)/permissions/page.test.tsx
@@ -416,19 +416,20 @@ describe("PermissionsPage", () => {
     expect(screen.getAllByText("npm-local").length).toBeGreaterThan(0);
   });
 
-  it("renders service account principal type with a human-readable label", async () => {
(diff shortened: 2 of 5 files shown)


brandonrc (changes requested):
Thanks for picking up #621 — this is a well-scoped fix and the test coverage is genuinely good: the payload-asserting unit test (`sends service_account in the create payload`), the missing-`principal_name` label-resolution test, and an E2E test that runs against the real seeded `e2e-ci-bot` service account (passed in shard 1) is exactly the bar we want.

I verified the API contract against the backend and everything checks out: `service_account` is accepted by `PermissionService::validate_principal`, the `svc-*` username prefix you search by matches `build_service_account_username`, and — importantly — the permissions list endpoint only resolves `principal_name` for `user` and `group` (the `CASE` in the list query has no service-account branch), so resolving labels client-side from the loaded service accounts is the correct approach, not a workaround. The named `serviceAccountsApi` import is also the right shape post-#694, and the unpaginated service-accounts endpoint means no truncation-notice concerns there.

Two things block merge, both fallout from this branch predating #688/#692/#694 — the same rebase situation as your #660:

**1. Rebase required — the PR is CONFLICTING with main (blocking).** There is exactly one conflict hunk, in `src/app/(app)/(admin)/permissions/page.tsx` in the query-declaration block. Main (#692) replaced the page's repository query with the shared hook:

```ts
const { data: repositoriesData } = useRepositories(
  { per_page: 1000 },
  { enabled: !!currentUser?.is_admin },
);
```

Resolution: keep main's `useRepositories(...)` call, move your new `service-accounts` `useQuery` next to it, and delete the hand-rolled query at page.tsx:152-157 plus the now-unused `repositoriesApi` import at page.tsx:17. `src/lib/api/permissions.ts` merges cleanly with #694 (your type-set change and the `unwrap()` refactor don't overlap), so no action needed there — just confirm the build after rebasing.

**2. Don't reintroduce the ad-hoc `["admin-repositories"]` cache key (convention, blocking).** The `useQuery({ queryKey: ["admin-repositories"], queryFn: () => repositoriesApi.list(...) })` at page.tsx:152-157 is precisely the pattern #692 (`useRepositories` / `repositoryListKey`) removed repo-wide: it bypasses prefix invalidation under `QUERY_KEYS.REPOSITORIES`, so this picker can serve a stale repository list after repos are created/renamed elsewhere. The rebase resolution above fixes this too — flagging it explicitly so it doesn't survive a mechanical conflict resolution.

Non-blocking, worth fixing while you're in there: `principalPickerOpen` / `principalSearch` (page.tsx:124-125) are never reset when the create dialog closes. If the user cancels with the popover open or a search typed, reopening the dialog restores a stale open popover and stale filter (Copilot flagged this as well). Resetting both in the dialog's `onOpenChange` at page.tsx:662 alongside `setCreateOpen(o)` is enough.

CI note: the shard-3 failure is `maven-grouped-browser.spec.ts` — unrelated to this change and green on main, so a re-run after the rebase should clear it. Everything else (lint, unit, build, shards 1-2, RBAC, visual) is green.

Once rebased with the query swapped to `useRepositories`, this is good to merge.

brandonrc:
Hi @ivolnistov — thanks for this fix; the implementation and test coverage were in great shape. Since we hadn't heard back and this is wanted for the 1.7.0 cut, a maintainer has adopted the branch and applied the review feedback (rebase onto main with the `useRepositories` resolution, plus the dialog picker-state reset) in a superseding PR: #710.

Your commits are preserved there with your authorship intact, and you're credited in the PR body. This PR will close automatically when #710 merges (or a maintainer will close it). Thanks again for picking up #621!

brandonrc:
Superseded by #710 (merged) — your commits landed with authorship preserved, plus the dialog-reset fix and coverage tests from the review. Thanks @ivolnistov!

PR #740, opened 2026-08-04
fix(repositories): show search matches from every page, not just the current one

### LINKED ISSUE #739: Repository search only matches repos on the current page
## Bug

On the Repositories page, typing in the search box only filters the repositories already loaded on the current page. If a repository whose name/key matches lives on another page, it never appears in the results.

Separately, pagination stayed visible while searching even when the filtered result set fit on a single page.

## Expected

Searching should find matching repositories regardless of which page they're on, and pagination should reflect the filtered result set.

## Fix

Send the search query to the backend list endpoint (`q`) so the search runs server-side across all repositories. `total_pages` then reflects the filtered set, so pagination hides for a handful of matches and stays for many.

nicola-preda:
## Summary
Repository search only filtered the repositories already loaded on the current page, so a name/key match on another page never showed up. This passes the query to the backend list endpoint (`q`) so the search runs server-side across all repositories. Because `total_pages` now reflects the filtered set, pagination auto-hides for a handful of matches and stays available when there are many.

Fixes #739

## Test Checklist
- [ ] Manually tested locally

## UI Changes
- [ ] Dark mode verified
- [x] N/A - no UI changes

### DIFF
--- a/src/app/(app)/repositories/_components/repositories-content.tsx
+++ b/src/app/(app)/repositories/_components/repositories-content.tsx
@@ -56,6 +56,9 @@ export function RepositoriesContent() {
   const [formatFilter, setFormatFilter] = useState("__all__");
   const [typeFilter, setTypeFilter] = useState("__all__");
   const [searchQuery, setSearchQuery] = useState("");
+  // Defer the search value so the backend list query (`q`) and artifact search
+  // don't refetch on every keystroke.
+  const deferredSearch = useDeferredValue(searchQuery);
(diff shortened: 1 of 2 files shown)


nicola-preda:
@brandonrc When using the search function on the repositories page with a lot of repos (i have 90) the pagination is still active, and results from the second page are not filtered. This PR fixes it (CI failure unrelated)

brandonrc (approved):

Issue #781, opened 2026-08-10
ci: visual regression job regenerates baselines instead of comparing them


brandonrc:
## Problem

The `E2E Visual Regression` job cannot fail on a visual diff, so it provides no regression coverage today.

Two independent causes:

**1. The run always rewrites baselines.** `.github/workflows/ci.yml` runs the suite with `--update-snapshots`:

```
npx playwright test --project=visual --update-snapshots --reporter=list,github,html
```

With `--update-snapshots`, Playwright writes the observed screenshot over the baseline rather than comparing against it. The job therefore reports success regardless of what the pages actually render.

**2. Every committed baseline is for the wrong platform.** All 31 committed baselines under `e2e/suites/visual/` are named `*-visual-darwin.png`. CI runs on `ubuntu-latest`, where Playwright looks for `*-visual-linux.png`. There are zero Linux baselines in the repo:

```
$ find . -name "*-darwin.png" | wc -l
31
$ find . -name "*-linux.png" | wc -l
0
```

So even without `--update-snapshots` the job would generate all 31 baselines from scratch on first run rather than compare against anything.

## Why it matters

The job's green status implies visual regressions are being caught. They are not. This surfaced during #752 / #777, where the new aggregate `E2E` check was justified partly on visual regression catching dependency-bump breakage. That justification does not hold while this is true, and the PR body was corrected accordingly.

## Suggested fix

- Drop `--update-snapshots` from the CI invocation so the suite compares.
- Generate and commit Linux baselines, ideally from the same container CI uses so font rendering matches.
- Consider putting baseline regeneration behind a manually dispatched job or a label rather than running it on every build.
- Decide whether the darwin baselines stay committed for local development or are gitignored.

Worth expecting real diffs on the first comparing run: 31 pages of never-verified baselines may surface genuine problems.

PR #785, opened 2026-08-13
fix: bump package.json to 1.8.0 and gate version against the release tag

### LINKED ISSUE #784: package.json still reads 1.7.0 on the v1.8.0 tag: web 1.8.0 reports itself as 1.7.0 in the UI
## Summary

`package.json` was not bumped for the 1.8.0 release, so a 1.8.0 build reports itself as **1.7.0** in the UI.

## Evidence

`package.json` at tag `v1.8.0`:

```json
{
  "name": "artifact-keeper-web",
  "version": "1.7.0",
```

`main` currently reads `1.7.0` as well. Release "Artifact Keeper Web 1.8.0" (tag `v1.8.0`) published 2026-08-03, and is the Latest release.

## Why it shows up in the UI

The sidebar renders two versions side by side — Web `NEXT_PUBLIC_APP_VERSION` and Server `health.version`. The Server half flows live from the backend. The Web half is baked into the bundle at build time, and `next.config.ts` inlines it from `pkg.version`. So an operator on web 1.8.0 sees "Web 1.7.0", which is not a version that was ever released as this build.

There are two sources for that number and they can disagree: the committed `package.json`, and the `APP_VERSION` build arg that CI passes from the git tag (`Dockerfile` `ARG APP_VERSION=dev` → `ENV NEXT_PUBLIC_APP_VERSION=${APP_VERSION}`). Which value wins depends on how the image was built, so the displayed version is currently ambiguous rather than merely wrong — worth confirming which one the published 1.8.0 image actually carries.

## Why it matters more than a cosmetic label

The displayed version is what a user quotes in a bug report and what an operator checks when deciding whether a fix has landed. A build that under-reports its own version causes a fix present in 1.8.0 to be diagnosed as missing, and sends whoever is investigating to the wrong changelog.

This is a recurring defect in this repo rather than a one-off — the same decoupled-source problem produced a stale UI version before. That history is the argument for a check rather than another manual fix.

## Requested

1. Bump `package.json` to match the released version, and confirm what the published 1.8.0 image actually reports so it is clear whether the build arg was masking this.
2. Add a release-time assertion that `package.json`'s version equals the tag being cut. The backend repo hit the identical class of bug through a hardcoded version in its OpenAPI spec; a one-line gate in the tag workflow is cheap and permanent, and manual discipline has now failed here more than once.

brandonrc:
## Summary

`package.json` was not bumped for the 1.8.0 release, so the published 1.8.0 build reports itself as **1.7.0** wherever the app names itself.

## What the published 1.8.0 image actually reports — measured, not inferred

The issue correctly flagged that two sources could disagree and that the answer should be established before writing a fix. It was, by pulling `ghcr.io/artifact-keeper/artifact-keeper-web:1.8.0` and reading the built bundle.

**The image reports `1.7.0`.** The client chunk carries the literal at the sidebar render site:

```js
children:["Web ","1.7.0","1.7.0".includes("-")&&1?` (${"aa02f04d..."})`:...]
```

`"1.7.0"` appears 4 times across the static chunks (sidebar + the admin Settings "Web Version" row). `1.8.0` appears **zero** times anywhere in `.next`. The image's own `/app/package.json` also reads `1.7.0`.

**The build arg did not mask it, and cannot.** CI passes `APP_VERSION=${{ github.ref_name }}` — literally `v1.8.0` — into `ENV NEXT_PUBLIC_APP_VERSION`. It has no effect, because `next.config.ts` sets the same variable to `pkg.version` literally:

```ts
env: {
  NEXT_PUBLIC_APP_VERSION: pkg.version,   // wins
  NEXT_PUBLIC_GIT_SHA: getGitSha(),       // reads process.env.GIT_SHA explicitly
}
```

The explicit `env` config value takes precedence over the environment variable of the same name. `GIT_SHA` is the control that confirms the mechanism rather than just the symptom: the bundle carries the correct commit SHA (`aa02f04d…`, the real `v1.8.0` commit) because `next.config.ts` reads `process.env.GIT_SHA` on purpose — so build args do reach the build, and this specific one is overridden.

Two consequences worth being explicit about:

1. **The user-visible impact is exactly what the issue assumed** — no ambiguity, and not cosmetic. The version is what a user quotes in a bug report and what an operator checks to decide whether a fix landed, so every fix in 1.8.0 can be diagnosed as missing.
2. **`package.json` is the sole source of the displayed version**, regardless of how the image is built. That makes gating `package.json` both necessary *and* sufficient — a gate on the build arg would have been theatre.

Incidentally, had the build arg won, it would have rendered `Web v1.8.0` with the leading `v`, since `github.ref_name` is the raw tag.

## Changes

1. **`package.json` → `1.8.0`** on main, plus the two matching `package-lock.json` entries (root and `packages.""`) so the lockfile stays consistent. Nothing else in the lockfile is touched.
2. **`scripts/assert-version-matches-tag.sh`** — asserts `package.json`'s version equals the tag being cut. Accepts `refs/tags/v1.8.0`, `v1.8.0`, or `1.8.0`.
3. **Wired into both tag-triggered workflows** so a mismatch cannot ship:
   - `docker-publish.yml`: a `version-check` job that `build` depends on, so no image is published.
   - `release.yml`: before the Release is created.
4. **A note on the Dockerfile `ARG APP_VERSION`** recording that it does not reach the bundle, so nobody reaches for it to correct a wrong version.

## Verified against the real defect

Restoring the pre-fix `package.json` and running the gate with the tag that was actually cut:

```
$ ./scripts/assert-version-matches-tag.sh v1.8.0
::error::package.json version does not match the tag being cut

  tag being cut     : v1.8.0  (expected version 1.8.0)
  package.json says : 1.7.0
exit=1
```

So it would have blocked this release. It passes on this branch, and on the `1.8.0` / `refs/tags/v1.8.0` argument forms.

## One design detail worth flagging in review

The `version-check` job is unconditional and the **assertion step** carries the `if: startsWith(github.ref, 'refs/tags/v')`, rather than putting the `if` on the job. A job-level `if` that evaluates false is *skipped*, and a skipped job skips everything that `needs` it — which would have silently disabled the Docker build on every branch push. This ordering keeps the job a fast no-op on branches while remaining a real dependency.

## Trade-off

The gate runs at tag time, so it stops a mismatched release rather than preventing the mismatch from reaching `main`. Catching it earlier would require deciding on `main` what the next version will be, which this repo deliberately does not do. Stopping the cut is the point at which the information is unambiguous.

Related: the backend repo has the same bug class through a hardcoded version in `openapi.rs`. The same one-file assertion would apply there; that is a separate repo and out of scope here.

Fixes #784


brandonrc:
Adjacent finding while checking this change against the visual suite — not fixed here, flagging for a follow-up.

`e2e/visual-mask.css` carries a rule intended to keep version numbers out of visual baselines:

```css
/* Version numbers in footer/header */
[data-testid="app-version"] {
  visibility: hidden !important;
}
```

`data-testid="app-version"` does not appear anywhere in `src/`, so that selector matches nothing and the rule is dead. Both places the app renders its version — the sidebar (`app-sidebar.tsx`) and the admin Settings "Web Version" row — are plain unlabelled elements, and both are inside committed baselines (`sidebar-expanded-visual-linux.png`, `settings-desktop-admin-visual-linux.png`). So every version bump nudges those two baselines even though a rule exists specifically to prevent it.

**This PR is safe against it.** The sidebar baseline is 256x991 = 253,696 px and the test allows `maxDiffPixelRatio: 0.01`, a budget of ~2,536 px. Changing `1.7.0` to `1.8.0` alters one digit glyph — on the order of tens of pixels, roughly two orders of magnitude inside the budget.

Worth noting it only became observable now: until #781 the visual job ran with `--update-snapshots` and rewrote baselines instead of comparing, so a version-driven diff would have been silently absorbed. Now that the job genuinely compares, the dead mask is a real (if currently small) coupling between a version bump and the visual suite.

Adding `data-testid="app-version"` to both render sites would make the existing rule do its job. Kept out of this PR to hold scope to the version defect.

PR #791, opened 2026-08-15
Complete localization with next-intl for English and Chinese UI

### LINKED ISSUE #792: Add Simplified Chinese (zh) localization for the web UI
### Problem Statement

The web UI is currently English-only. Chinese-speaking users are forced to
navigate an English-only interface, and there is no way to switch languages.
Because all UI text is hardcoded in English at the component level, localizing
any part of the app (or keeping it in sync across locales) requires touching
individual components — there is no central catalog, no fallback, and no
enforcement that new strings get translated.

This affects the whole surface: pages, dialogs, the auth flow, error screens,
shared components, and even the browser-tab title/description, which stay in
English regardless of the user's language.

### Proposed Solution

Add full localization to the web UI with next-intl:

- Resolve the locale per request from a `NEXT_LOCALE` cookie — no URL prefix —
  so existing deep links, bookmarks and Playwright E2E specs keep working
  unchanged.
- Ship English (`en`) and Simplified Chinese (`zh`), with a language switcher
  in the app header and on the login page.
- Store message catalogs as flat JSON files in `src/i18n/locales/{en,zh}/`
  mirroring the source tree (one `page.json` per `page.tsx`, one `*.json` per
  component), with a file's namespace equal to its locale-relative path.
- Load messages server-side per route directory; nested providers compose the
  shared `CORE_ROOTS` with each route's own directory, so a page only ships the
  messages its subtree renders.
- Fall back to English for any untranslated key so a raw message key never
  leaks to the UI.
- Enforce quality with a check script (`scripts/check-i18n.mjs`: JSON-only
  catalogs, en/zh key parity, every file loaded by some layout) and a global
  next-intl test mock so translated components stay testable.

Expected outcome: the entire UI is usable in Simplified Chinese, and adding a
third language later is a matter of mirroring the `en/` tree and registering
the locale code — no per-component changes.

### Alternatives Considered

_No response_

### Use Case

_No response_

### Component

None

### Additional Context

_No response_

### Pre-submission Checklist

kaicen-x:
## Summary

This PR bundles the user-requested follow-up work since the 1.8.0 version-gate
commit (#785): full Simplified-Chinese localization, a brand default theme,
and the supporting docs/build fixes.

### 1. Localization (i18n) — the UI now ships in English and Simplified Chinese
- Added a next-intl framework; the locale is resolved per request from the
  `NEXT_LOCALE` cookie (no URL prefix), so existing deep links, bookmarks and
  E2E specs work unchanged. A language switcher appears in the app header and
  on the login page.
- Translated every page and shared component into `zh`: the admin console,
  protected pages, app pages, the auth flow, shared components, the
  404/error/global-error screens, and the browser-tab title/description
  (via `generateMetadata`).
- Restructured the catalogs: `src/i18n/locales/{en,zh}/` holds only flat JSON
  files mirroring the source tree — one `page.json` per `page.tsx`, one
  `*.json` per component; a file's namespace is exactly its locale-relative
  path. `loadMessages` (server-only) reads the JSON from disk and loads by
  route directory: no groups, no aggregators, no namespace lists. Nested
  providers compose the shared `CORE_ROOTS` with each route's own directory.
- `scripts/check-i18n.mjs` enforces JSON-only catalogs, en/zh key parity, and
  that every message file is loaded by some layout; a global next-intl vitest
  mock keeps translated components testable.
- Follow-up fixes included: setup-dialog header missing-key guard, duplicate-
  key override fix, per-component namespaces (incl. `_components/`), the npm
  upstream change-feed card's missing env-var descriptions, and translated
  repository-type choices in the create-repo dialog.

### 2. Brand default theme
- `brand` becomes a standalone next-themes theme (default) — white/navy
  palette (`#023795` primary, `#4690d2` secondary), white top header, light-
  blue page background with a soft glow gradient scoped to `.brand`.
- The original `light` and `dark` themes are left byte-for-byte unchanged; the
  header toggle now cycles brand → light → dark.

### 3. Docs & build
- Refreshed README / README-zh (current stack, i18n and theme sections).
- Added `@swc/helpers` 0.5.23 as a direct dependency so `npm ci` in the Docker
  build satisfies `@swc/core`'s peer (image builds had been failing with
  "Missing: @swc/helpers@0.5.23 from lock file").

### Issues
- Closes #792

## Test Checklist

- [x] Manually tested locally

## UI Changes

- [x] Dark mode verified
- [ ] N/A - no UI changes


brandonrc:
Thank you for the amount of work here. A Chinese UI is a reasonable goal, and the zh translations I sampled read well. I can't merge this PR as it stands, and it's too large and too far behind `main` for a maintainer to take over. Here is why, and what would make it landable.

**Why not as-is**
- **No maintainer decision yet.** Runtime i18n changes how every future UI PR is written, since every string goes into two catalogs. The project hasn't decided to adopt it, and #792 has no maintainer discussion yet.
- **The default theme changes for everyone.** Main defaults to `system` (it follows the OS dark mode). This PR makes a new `brand` theme the default and removes `system`. That needs its own issue and decision.
- **English copy changes silently**, although the rendered text is described as unchanged. SSO row buttons lose the provider type from their accessible names ("Edit OIDC provider X" → "Edit X"). "Upload size limit saved" becomes "Max upload size saved". "S3" becomes "S3 (AWS)". Lowercase enum values are title-cased. Unit and E2E assertions were edited to match.
- **One message will not parse.** `settings/sso/page.json` → `userFilterHint` contains `{"{0}"}`, copied from JSX. ICU rejects it (`MALFORMED_ARGUMENT`). The test mock substitutes with a regex, so unit tests don't catch it.
- **Too far behind main.** Main is 27 commits ahead, covering the whole 1.10.0 milestone. 23 files conflict, and about 22 new components on main are untranslated.
- **Visual regression fails on the header.** The language switcher and theme icon changed it, and the baseline wasn't refreshed.

**What would be landable**
1. A short proposal on #792 first: whether to adopt runtime i18n, with the conventions written into ARCHITECTURE.md and AGENTS.md.
2. **PR A, rebased on current main:** the next-intl framework plus English catalogs only. No English copy changes, the 1.10.x components covered, and accessible names unchanged.
3. **PR B:** zh catalogs, the language switcher, and a refreshed header baseline.
4. The theme proposal as its own issue, keeping `system` as the default unless decided otherwise. Send the README rewrite on its own too.
5. Make `scripts/check-i18n.mjs` parse every message with the ICU parser and compare placeholders across locales, and run it in CI. Use real `intl-messageformat` in the test mock. Mark `load-messages.ts` as `server-only`.
6. Drop the direct `@swc/helpers` dependency. Current main (Next 16.3.x) already hoists 0.5.23, so the Docker issue it worked around no longer exists.
7. Use one CHANGELOG bullet per PR citing #792.

I'm leaving this PR open so the discussion stays in one place. Thanks again.

PR #794, opened 2026-08-15
feat(setup): add vscode gateway onboarding for VSCodium, code-server, and VS Code

### LINKED ISSUE #793: Setup Guide has no vscode onboarding for the Open VSX gallery gateway
Backend artifact-keeper/artifact-keeper#3253 (closed artifact-keeper/artifact-keeper#3252) shipped Artifact Keeper's VS Code/Open VSX extension gallery gateway in 1.8.0: VSCodium, code-server, and, with an enterprise policy, official VS Code can pull extensions through a governed AK repository instead of reaching open-vsx.org directly.

The PR's client-config doc was deleted before merge (`fae02f96`), so the shipped feature has no user-facing onboarding anywhere. Two gaps in the web UI:

The Setup Guide's vscode format has no case in `getRepoSetupSteps` (`src/components/setup/repo-setup-guide.tsx`), so opening the Setup Guide for a vscode repository shows the generic `curl -X PUT .../artifacts/` upload snippet, which is not how any of the three supported clients talk to a gallery.

The create-repository dialog has no default upstream URL for `vscode` in `src/app/(app)/repositories/_lib/default-upstream-urls.ts`, so an admin creating a new gateway repository gets no `https://open-vsx.org/vscode/gallery` suggestion the way every other Remote format does.

PR incoming with Setup Guide coverage for all three supported clients (VSCodium's persistent product.json, code-server's EXTENSIONS_GALLERY, and the official VS Code enterprise-policy constraint) plus the default upstream URL.

cazlo:
Title: feat(setup): add vscode gateway onboarding for VSCodium, code-server, and VS Code

Fixes #793

Backend artifact-keeper/artifact-keeper#3253 shipped Artifact Keeper's VS Code/Open VSX gallery gateway in 1.8.0, but the PR's client-config doc was deleted before merge, so the feature had no user-facing onboarding anywhere. The vscode format also had no case in `getRepoSetupSteps`, so the Setup Guide fell through to the generic curl-upload snippet, and the create-repository dialog had no default upstream URL for a new vscode Remote.

This adds a vscode case covering the three clients the backend supports:

VSCodium's persistent `product.json` `extensionsGallery` block (`serviceUrl`/`itemUrl`/`extensionUrlTemplate`/`latestUrlTemplate`/`controlUrl`), with a callout that `latestUrlTemplate` is required alongside `extensionUrlTemplate` — omitting it can let an update lookup fall back to open-vsx.org directly, bypassing the gateway.

code-server's `EXTENSIONS_GALLERY` env block.

Official VS Code's enterprise-policy constraint: desktop only, requires the `ExtensionGalleryServiceUrl` policy, VS Code 1.99+, an entitled signed-in account, and a supported policy deployment. I stated this as a real constraint rather than implying it works out of the box — it does not cover VS Code Server or VS Code for the Web either.

`default-upstream-urls.ts` gains `vscode: "https://open-vsx.org/vscode/gallery"`. I verified this against the merged backend source (`backend/src/api/handlers/vscode.rs`) rather than just the plan: the gateway validates that a Remote repository's `upstream_url` ends in `/vscode/gallery` and returns a 502 otherwise; `open-vsx.org` is the real public Open VSX instance and its adapter root satisfies that check. A live release-gate test in a separate unit also created a repository with this exact URL successfully against the merged backend.

Tests extend the existing colocated files rather than adding new ones: `repo-setup-guide.test.tsx` gets a flat-step-list rendering check (no tabs, all three client snippets present), and `setup/page.test.tsx` gets a dedicated describe block covering key interpolation into every URL, the `latestUrlTemplate` requirement and warning copy, the code-server env block, and the VS Code enterprise-policy text — plus `vscode` added to the existing non-JVM-formats `it.each` table. `default-upstream-urls.test.ts` gets a dedicated assertion that the value is the gallery adapter root, not the Open VSX site root or its native `/api` path.

Local validation: `npm run lint` clean (0 errors, pre-existing warnings only, none in touched files), `npx tsc --noEmit` shows the same 22 pre-existing errors in unrelated files (`audit`/`blast-radius` test fixtures) whether or not this change is applied, and `npm run test` passes all 174 files / 3270 tests, including the new/extended cases in the three touched test files.

---
---

## Note on the visual-regression commits (now removed)

This PR briefly carried a second commit stubbing the monitoring page's API responses, because `monitoring - desktop` failed on every PR opened against `main` and this one could not go green without it.

#838 has since fixed that race upstream, and better: the E2E seed records a health check via `POST /api/v1/admin/monitoring/check` so the populated layout is guaranteed before any screenshot, rather than one branch of the race being pinned. With the regenerated 720px baseline my stub inverted the failure — it forced the 918px empty state — so I dropped it. `e2e/suites/visual/pages/admin-pages.spec.ts` is now identical to `main`.

The net diff of this PR is the vscode onboarding only: the Setup Guide case, the default upstream URL, their tests, and one CHANGELOG entry.


brandonrc:
Thanks for this. I checked every URL in the three snippets against the backend's `vscode.rs` routes, and they all resolve. The `/vscode/gallery` suffix requirement for the default upstream is also correct.

There is one behaviour mismatch. The gallery routes return **501 for non-Remote** repositories and **403 for non-public** ones (`gallery_upstream()`), but the new `case "vscode"` shows the gallery snippets for every vscode repository. Here is what I'll push to this branch, keeping your commits and authorship:
1. For a vscode repository that is not a public Remote, the Setup Guide shows one explanatory step instead: the gallery only serves public Remotes, and the direct `…/extensions/<publisher>/<name>/<version>/download` route is available. The three gallery steps stay for public Remotes. I'll add tests for the local and private-remote cases.
2. Rebase onto current `main`. Git merges it without a conflict, but the anchor lines of your CHANGELOG bullet now sit under the released `[1.10.0]` heading. I'll move the bullet to `[Unreleased] / ### Added`.

I'll squash-merge once CI is green.

PR #813, opened 2026-08-24
feat(auth): silent SSO auto-login (check-sso) with fail-open iframe probe

### LINKED ISSUE #812: Silent SSO auto-login: sign in users with a live IdP session automatically, keep anonymous visitors anonymous
## Problem

With an OIDC provider configured, a user who already holds a live Keycloak/IdP session still has to click **Sign in with Keycloak** on every fresh visit. Meanwhile a naive sole-provider auto-redirect is unacceptable: registry browsing is legitimately anonymous (public repos, proxy indexes), and anonymous visitors must never be bounced to a visible IdP login page uninvited.

## Proposal (check-sso)

On app load, when nobody is signed in, an enabled OIDC provider exists, the operator has not disabled it, and no prior attempt is recorded for this browser session:

- probe the IdP once with `prompt=none` (backend artifact-keeper#3546) **inside a hidden same-origin iframe** — never a top-level redirect;
- live IdP session → the normal callback signs the user in invisibly (cookies set, identity adopted, queries refetched);
- no session → the backend relays the IdP's `login_required` as `/callback?silent_denied=1`; the page stays anonymous with **no error UI**;
- IdP down/slow → the probe times out quietly (fail open): anonymous browsing continues undisturbed.

One attempt per browser session (sessionStorage guard), cleared on explicit sign-in and sign-out. The explicit "Sign in with …" button behavior is unchanged. Operator kill switch: backend `OIDC_SILENT_SSO=false`, surfaced as `auth.silent_sso_enabled` in the system config the app already fetches.

Requires relaxing `frame-ancestors`/`X-Frame-Options` from deny to **same-origin only** on exactly the `/callback` + `/auth/callback` routes so the probe iframe can land there; everything else stays unframable.

brandonrc:
## Summary

Closes #812. Web half of silent SSO auto-login; pairs with backend PR artifact-keeper/artifact-keeper#3547 (which adds `prompt=none` support, the `silent_denied` callback contract, and the `OIDC_SILENT_SSO` kill switch).

Users who already hold a live IdP session are signed in invisibly on page load; everyone else keeps browsing anonymously with zero visible change.

**How it works.** `SilentSsoBootstrap` (mounted app-wide, renders nothing) runs at most one probe per browser session: when the auth bootstrap finishes with nobody signed in, the system config allows it, and an enabled OIDC provider exists, it makes one silent `prompt=none` round trip through the IdP. A live IdP session flows through the normal backend callback (cookies set on the 307), and the callback page returns the user to the page they started on, signed in. No session → the IdP answers `login_required` → the backend relays it as `/callback?silent_denied=1` → the callback page returns the visitor to where they were, still anonymous, with no error UI.

### The three objections, addressed head-on

- **Anonymous stays anonymous.** Registry browsing is legitimately anonymous, so this is deliberately *check-sso*, not a sole-provider auto-redirect: `prompt=none` forbids the IdP from rendering any UI, so an anonymous visitor is never bounced to a visible Keycloak page — the round trip is a pair of instant redirects — and the denied probe produces no error UI.
- **IdP down ⇒ fail open.** Before navigating anywhere, a short-timeout (4s) same-origin **preflight** fetches the `prompt=none` login URL with redirects disabled. The backend only answers with a redirect after its *own server-side fetch of the IdP discovery document succeeded*, so this one cheap request vouches for both the backend and the IdP; timeout / network error / 4xx / 5xx ⇒ no navigation, visitor undisturbed. Exactly one attempt per session — the sessionStorage guard is written *before* the redirect so an interrupted round trip can never loop, and storage-unavailable browsers fail closed to "already attempted". The return path is stored in sessionStorage and validated on consumption as a same-app path (never an open redirect).
- **Kill switch.** Backend `OIDC_SILENT_SSO=false` surfaces as `auth.silent_sso_enabled` in the system config the app already fetches; the bootstrap then never starts. Default on, including against older backends that omit the field (there the preflight fails invisibly once — pre-flag backends reject `prompt` — and the guard stops retries).

### Why a top-level redirect and not a hidden iframe

An earlier revision of this branch used a hidden same-origin iframe (keycloak-js `silentCheckSsoRedirectUri` style). Live verification against a real Keycloak killed it: **IdP session cookies are frequently `SameSite=Lax`** (Keycloak issues Lax unless it is certain the deployment is HTTPS-proxied), and Lax cookies are never sent on a cross-site iframe navigation — the iframe probe silently reported "no session" for users who had one, on exactly the deployments this feature targets. A top-level navigation carries Lax cookies, so it works on both Lax and None IdPs, at the cost of one invisible page round trip per session. This also means **zero security-header changes**: the frame-ancestors / X-Frame-Options / frame-src relaxations from the earlier revision are fully reverted.

**Unchanged:** the explicit "Sign in with …" button (no `prompt` param), `?fallback=local`, LDAP and local login, demo auto-login, first-boot setup, all security headers. The guard clears on explicit sign-in and sign-out, so sign-out followed by a reload can silently re-sign-in while the IdP session lives — the standard check-sso semantic (AK sign-out is not IdP sign-out). No probe starts on `/login`, `/callback`, or `/change-password`.

## Testing

- `npm run lint`: 0 errors
- `npm run test`: 3335 passed (new: silent-sso lib incl. preflight fail-open matrix — opaqueredirect/500/network/timeout — and open-redirect rejection on the return path; bootstrap orchestration incl. kill switch, guard, excluded paths, SAML-only skip; callback-page silent_denied / return-to / exchange contracts; system-config parsing)
- `npm run build`: production build succeeds

Live verification on a real Keycloak deployment (patched images): with an IdP session, visiting / signs the user in with no click and returns them to the page they were on; a fresh session stays anonymous with exactly one probe, no redirect loop, and no error UI; login page + `?fallback=local` intact; explicit SSO login, admin group mapping, and anonymous package-index access unaffected; a blocked/hung login endpoint leaves the visitor browsing anonymously (preflight timeout).


brandonrc:
CI status note: every check is green except **E2E Visual Regression**, which fails on `e2e/suites/visual/pages/admin-pages.spec.ts › monitoring - desktop` — the same screenshot comparison that fails on the current tip of `main` (see the check runs on 6c9986d). This PR does not touch the monitoring page; the failure pre-dates the branch and reproduced identically on a rerun (it passed once on an earlier commit of this branch, so it is flaky/stale-baseline rather than deterministic). Leaving the PR unmerged until the baseline is fixed on main or a maintainer confirms the check can be waived.

brandonrc:
## Review — NEEDS-WORK

Reviewed at `63ac064` with the backend contract (#3547) side by side; `npm ci`, `tsc` (0 errors), lint (0 errors), vitest 3335 pass, `next build` OK, merge from main clean.

**First, a naming correction:** the title, #812 and the backend PR body all describe a hidden iframe probe, but this code does a preflight-gated **top-level redirect**. No iframe, no `postMessage`, so no origin check is needed and `frame-ancestors 'none'` / `X-Frame-Options: DENY` stay as they are. Fail-open means anonymous, never authenticated. The `silent_denied` contract matches the backend exactly (`sso.rs:299` ↔ `callback/page.tsx:42`).

**Blockers**

1. **Redirect loop.** `callback/page.tsx:104` clears the once-per-session guard *before* `refreshUser()`, which swallows errors and leaves `user = null`. `SilentSsoBootstrap` stays mounted across the client-side `router.replace` with `startedRef` still false, so it probes again on the next render, forever, whenever `/me` keeps failing (Secure cookie over plain HTTP, domain mismatch, 5xx). Deleting that line is enough; `auth-provider.tsx:88` already clears on success.
2. **Sign-out immediately signs back in.** `logout()` clears the guard (`auth-provider.tsx:163`), the header pushes to `/` (`app-header.tsx:47-48`), and the bootstrap effect fires on `user → null` on that same page. Artifact Keeper performs no IdP end-session, so an OIDC user cannot sign out while the Keycloak session lives. This contradicts #812's "clear on sign-out" requirement, which should be revisited rather than implemented.

**Should-fix:** no timeout on the navigation itself (the preflight only vouches for discovery); "once per session" is really once per tab; each probe leaves two orphan `sso_sessions` rows.

**Nits:** `login_url.startsWith("/")` admits `//host` (`silent-sso-bootstrap.tsx:65`); `/error/*` is not excluded; the "hidden iframe" wording persists in the title and `system-config.ts:47-49`.

**Bookkeeping:** the CHANGELOG entry follows AGENTS.md but should note the backend requirement — #3546/#3547 shipped in backend **1.8.2** (2026-09-01), not 1.9.0. Tests are missing exactly the two transition cases above.

PR #818, opened 2026-08-26
fix(replication): show the real replication mode instead of defaulting to pull

### LINKED ISSUE #817: Replication subscriptions table shows every assigned repo as Pull regardless of the stored mode
The **Subscriptions** tab of `/replication` shows a per-repository **Replication Mode** dropdown. For any repository assigned to the selected peer, that dropdown always reads **Pull**, whatever the subscription actually stores.

The value is fabricated in the page:

```tsx
const currentMode = repoModes[r.id] ?? (assignedSet.has(r.id) ? "pull" : "none");
```

`peersApi.getRepositories` calls `GET /api/v1/peers/{id}/repositories`, which returns a bare array of repository ids. With no mode in the response, the page has nothing to render and falls back to `"pull"` for every assigned repo.

### Impact

A `push` or `mirror` subscription is displayed as `pull`, and the display is confidently wrong rather than blank. Since the same dropdown is also the editing control, opening the tab and saving an unrelated row can write `pull` over a mode the operator never chose. Replication direction is not a detail an operator can afford to read wrong.

There is also no indication anywhere in the UI of which side "push" and "pull" are relative to.

### Fix

Backend `artifact-keeper#3575` makes the list endpoint return `Vec<SubscriptionResponse>`, the same type the single-subscription route already serves, so `replication_mode` arrives with the list. The frontend then renders the stored mode instead of guessing, and the column header gains a tooltip stating that direction is relative to this instance.

Depends on artifact-keeper/artifact-keeper#3575 (PR artifact-keeper/artifact-keeper#3576, must merge first).

nicola-preda:
## Summary

Fixes #817

**Depends on artifact-keeper/artifact-keeper#3576 (issue artifact-keeper/artifact-keeper#3575). That PR must merge, and the SDK must be regenerated, before this one is useful.**

The Subscriptions tab of `/replication` fabricated the replication mode it displayed:

```tsx
const currentMode = repoModes[r.id] ?? (assignedSet.has(r.id) ? "pull" : "none");
```

`GET /api/v1/peers/{id}/repositories` returned a bare array of repository ids, so the page had no mode to render and defaulted every assigned repository to `pull`. A `push` or `mirror` subscription displayed as `pull`, and since that dropdown is also the editing control, the wrong reading was one save away from becoming the stored value.

With artifact-keeper#3576 the list endpoint returns `Vec<SubscriptionResponse>`, so this consumes the real `replication_mode` and renders it. The column header also gains a tooltip stating that direction is relative to this instance (push = this instance sends the repo to the peer), which the UI stated nowhere before.

Two incidental changes the fix needs:

- `DataTableColumn.header` widens from `string` to `React.ReactNode` so the tooltip can live in the header. Every existing caller passes a string and is unaffected.
- `peersApi.getRepositories` reads the subscription shape through a passthrough cast, because the generated SDK still types the response as `string[]`. This follows the same approach `admin.ts` uses for the health `image_tag`, and the cast comes out when the SDK is regenerated against artifact-keeper#3576.

**Behaviour against a backend without artifact-keeper#3576:** the endpoint still returns bare ids, so the tab reads every repository as unassigned with mode `None`. That is visibly empty rather than confidently wrong, but it is a regression against today's behaviour, which is why the merge order matters. I am happy to add a shape-tolerant adapter that accepts both the id array and the subscription list if you would rather this be safe to merge in either order, or to hold this PR until the backend lands and the SDK ships. Your call on which you prefer.

## Test Checklist
- [ ] Manually tested locally

Being explicit about what is and is not covered here:

`getRepositories` is covered by an updated unit test in `src/lib/api/__tests__/replication.test.ts` asserting the subscription shape round-trips, including a `sync_enabled: false` row. The render path is **not** unit-covered: there is no `page.test.tsx` for `/replication` in the repo, so nothing asserts that the dropdown shows the stored mode rather than `pull`. Adding a page-level test for the Subscriptions tab is the obvious gap and I will write one if you want it in this PR.

E2E is unchanged. `e2e/suites/interactions/integrations/replication.spec.ts` is smoke-level (headings, tabs, peer-selector visibility) and asserting the mode column needs a live backend carrying artifact-keeper#3576. "Manually tested locally" is left unchecked for the same reason: the display cannot be exercised against a backend that does not return the field yet.

Run green: the 27 `src/lib/api/__tests__/replication.test.ts` tests, the 215 tests under `src/components/common` (for the `DataTableColumn` change), `tsc --noEmit`, and `eslint` on all four changed files.

## UI Changes
- [ ] Dark mode verified

The only new markup is a `Tooltip` in a table header, using the same `@/components/ui/tooltip` primitive already used further down this page, so responsive behaviour, dark mode and the tooltip's own keyboard and ARIA handling come from the shared component rather than anything new here. I have not verified them against a running instance, so I have not checked those boxes.

No visual baselines are affected: `e2e/suites/visual/pages/core-pages.spec.ts` screenshots `/`, `/repositories`, `/packages`, `/search` and `/login`, and `/replication` is not among them.


nicola-preda:
The red E2E here isn't from this PR. `monitoring - desktop` compares a 1280x918 baseline captured on web 1.7.0 / server 1.7.1, where the page was still empty ("No alert data"), against a 1280x720 render where the stack has recorded a health check and the page shows a service card plus a one-row Health Log. Mismatched dimensions fail before the pixel threshold applies. The dependabot PRs that only bump an action fail on the same test.

Worth flagging that a refreshed baseline won't hold either: that Health Log row bakes in a live timestamp and response time, so it needs a mask to be stable. Happy to open a separate issue for it if that's useful.

brandonrc:
The two reds are the `monitoring-desktop-admin` snapshot, which neither PR touches — an old baseline at a different viewport, and a refresh will not hold while the Health Log renders a live timestamp. Please do open the separate issue for it: mask the timestamp (or freeze the clock) and re-baseline at the current viewport, in one PR that only touches the snapshot. Don't chase it in these two.

Merge order between this and artifact-keeper#3578 / #3576 does not matter to me; the backend side is milestoned 1.10.0 and I will land them together.

brandonrc:
Taking this over to land it with the web 1.10.0 milestone (it is the web half of backend artifact-keeper/artifact-keeper#3575, which changed the peer-repositories response shape). I rebased the branch onto current main, which resolved the CHANGELOG conflict and picks up the refreshed visual baselines that were failing the components snapshots here; the code is unchanged. Pushed as faa44a99 to this branch. It will merge through the milestone queue once CI is green.

PR #738, opened 2026-08-03
feat(dashboard): include proxy cache in storage/artifact stats with hover breakdown

nicola-preda:
## Summary
The Storage Used and Artifacts cards now show local + remote combined, with a hover tooltip splitting Local vs Remote. Reads the new `proxy_*` fields from the stats API via passthrough (until the SDK types land), defaulting to 0 on older backends.

Fixes #737
Depends on artifact-keeper/artifact-keeper#3087 (backend must merge first).

## Test Checklist
- [x] Unit tests added/updated
- [ ] E2E Playwright tests added/updated
- [ ] Manually tested locally
- [x] No regressions in existing tests

## UI Changes
- [ ] Playwright E2E spec covers the change
- [ ] Responsive layout verified (mobile + desktop)
- [ ] Dark mode verified
- [x] Accessibility checked (keyboard navigation, screen reader)
- [ ] N/A - no UI changes

Issue #737, opened 2026-08-03
Dashboard storage/artifact stats ignore proxy cache


nicola-preda:
## Problem Statement
The dashboard "Storage Used" and "Artifacts" cards only count hosted (local) content. On instances that mostly proxy upstreams, both show 0 even though the proxy cache is using real storage.

## Proposed Solution
Show the combined local + remote total on each card, and reveal a Local vs Remote split on hover.

Depends on the backend exposing the per-bucket figures: artifact-keeper/artifact-keeper#3087 (must be merged first).

## Component
Dashboard

## Pre-submission Checklist
- [x] I have searched existing issues to make sure this is not a duplicate.
- [x] I have reviewed the documentation to confirm this feature does not already exist.

nicola-preda:
@brandonrc Small fix to have both the Local and Remote artifact number / storage usage viewed on the admin panel. 

Needs a small backend fix: https://github.com/artifact-keeper/artifact-keeper/issues/3087

brandonrc:
> @brandonrc Small fix to have both the Local and Remote artifact number / storage usage viewed on the admin panel. 
> 
> Needs a small backend fix: https://github.com/artifact-keeper/artifact-keeper/issues/3087

Thanks! Trying to get the milestones in line. Working a 1.7.1 release. I'll make sure this gets in there.

PR #820, opened 2026-08-26
feat(analytics): local+remote totals with a split toggle

### LINKED ISSUE #819: Analytics page reports hosted storage only, so remote repositories read as empty
The analytics page presents hosted storage as if it were all the storage there is.

Proxy-cached objects have no `artifacts` row, so `GET /analytics/storage/breakdown` and `GET /analytics/storage/trend` report only hosted figures. On an instance with pull-through remotes a Remote repository shows 0 artifacts and 0 bytes on the analytics page no matter how much it has cached, and the storage trend charts a line that ignores the proxy half entirely.

The dashboard already knows better. It shows a local/remote split with a hover breakdown (#737 / #738), reading the proxy fields `/admin/stats` exposes. So the two pages disagree about the same storage, and the one built for storage analysis is the one that is wrong.

### What this needs

The analytics page should show local and remote totals with a way to see the split, reusing the vocabulary the dashboard already established rather than inventing a second one. The hover-breakdown component currently lives inline in `dashboard-content.tsx`; it wants extracting so both pages share it.

### Dependency

**This depends on artifact-keeper/artifact-keeper#3577 (PR artifact-keeper/artifact-keeper#3578), which must merge first**, because the breakdown and trend responses do not carry the proxy fields until it lands. The SDK also needs regenerating against it before the fields are typed.

nicola-preda:
## Summary

Fixes #819

**Depends on artifact-keeper/artifact-keeper#3578 (issue artifact-keeper/artifact-keeper#3577) for the data**, and on the SDK being regenerated against it for the types. See the note on merge order at the end: unlike #818, this one is safe to merge in either order.

The analytics page reported hosted storage as if it were all the storage there is. Proxy-cached objects have no `artifacts` row, so a Remote repository showed 0 artifacts and 0 bytes however much it had cached, and the storage trend charted a line that ignored the proxy half. The dashboard already showed a local/remote split with a hover breakdown (#737 / #738), so the two pages disagreed about the same storage and the one built for storage analysis was the wrong one.

This brings the analytics page to the same vocabulary: local and remote totals with a split toggle, and the per-repository breakdown carrying both halves.

`StatBreakdown` was defined inline in `dashboard-content.tsx`. It moves to `src/components/common/stat-breakdown.tsx` so the analytics page uses the same component rather than a second implementation of the same hover card. `dashboard-content.tsx` now imports it and is otherwise unchanged.

The proxy fields are read through a passthrough cast defaulting to 0, the same approach `admin.ts` already uses for the `/admin/stats` proxy fields. The casts come out when the SDK is regenerated.

**Merge order.** The `?? 0` defaults mean that against a backend without artifact-keeper#3578 the proxy halves read as 0, which is exactly the current behaviour: the split toggle shows everything as local and the totals match what the page shows today. So this degrades to today's page rather than to a broken one, and it does not have to wait for the backend the way #818 does. It just does nothing useful until the backend lands.

## Test Checklist
- [x] Manually tested locally

`npx vitest run` is 3307 passed across 177 files. New tests cover the analytics page (`analytics/__tests__/page.test.tsx`) and the adapters that read the proxy fields (`lib/api/__tests__/analytics.test.ts`), including the older-payload path where the fields are absent and must default to 0.

`npx eslint` is clean on every changed file. `npx tsc --noEmit` reports 22 errors, all of them in `(admin)/audit/__tests__/page.test.tsx` and `(admin)/security/blast-radius/__tests__/page.test.tsx`, neither of which this PR touches; they are pre-existing on `main`.

## UI Changes
- [ ] Dark mode verified
- [ ] N/A - no UI changes

Being straight about these four rather than ticking them: this changes the analytics page and the extracted breakdown component, and I verified it through unit tests only. I have not run it in a browser, so responsive layout, dark mode and keyboard/screen-reader behaviour are unverified by me. The extracted `StatBreakdown` is byte-identical to the markup already shipping on the dashboard, so it inherits whatever those properties already were there; the analytics page's new toggle is the part that genuinely has not been checked. Happy to add a Playwright spec if you want one before merge.


brandonrc:
The two reds are the `monitoring-desktop-admin` snapshot, which neither PR touches — an old baseline at a different viewport, and a refresh will not hold while the Health Log renders a live timestamp. Please do open the separate issue for it: mask the timestamp (or freeze the clock) and re-baseline at the current viewport, in one PR that only touches the snapshot. Don't chase it in these two.

Merge order between this and artifact-keeper#3578 / #3576 does not matter to me; the backend side is milestoned 1.10.0 and I will land them together.

brandonrc:
Taking this over to land it with the web 1.10.0 milestone (backend artifact-keeper/artifact-keeper#3577 ships the proxy storage split this reads). Plan: rebase onto current main (clean, code unchanged), regenerate the Linux visual baseline for the analytics page that this change legitimately alters, and push the result to this branch. The components snapshots failing here are stale baselines that the rebase fixes on its own.

brandonrc:
Rebased onto current main and pushed as ff465159 (code unchanged). I regenerated the Linux visual baselines on the rebased branch against the backend dev image and the analytics snapshot came out identical, so the earlier failure there was the stale-baseline problem too; no baseline change was needed. It will merge through the milestone queue once CI is green.

PR #824, opened 2026-09-01
fix(permissions): stop offering service accounts as User principals (#823)

### LINKED ISSUE #823: Permissions page: service accounts are listed under principal type "User", where selecting one always fails validation
### Description

On the admin **Permissions** page, the *Principal* picker for principal type **User** lists service accounts alongside real users. Selecting one is a guaranteed `400` from the API — `principal_id <uuid> does not exist as a user` — because the backend requires `principal_type: "service_account"` for those rows.

The page already holds everything needed to prevent this. It fetches the service-account list for its own labelling:

- `src/app/(app)/(admin)/permissions/page.tsx:151` — `serviceAccountsData` query
- `src/app/(app)/(admin)/permissions/page.tsx:193` — `serviceAccountNames`, an id → name `Map` built from it, used to label existing permission rows

But the `user` branch of `principalOptions` (`page.tsx:165`) maps `users` straight through without excluding those ids.

There is a backend half to this too — `GET /api/v1/users` returns service accounts with no `is_service_account` field and no filter parameter, which is why the other four screens consuming that list can't fix it locally. Filed separately as artifact-keeper#3634. **This issue is only about the Permissions page**, which is fixable today without waiting on the API.

### Environment Information

- Browser: Chrome
- Browser Version: current stable
- Operating System: macOS
- Screen Resolution: n/a

### Web UI Version

v1.8.0 (reproduced against `v1.8.0-38-g39ceb05`)

### Backend Version

v1.8.1

### Expected Behavior

The picker never offers a choice the API is certain to reject. Either the **User** list excludes service accounts (they are reachable under the **Service Account** type, which the form already supports), or picking one transparently switches `principal_type` to `service_account` instead of erroring.

### Observed Behavior

The service account appears under **User**. Submitting the form fails validation with `principal_id <uuid> does not exist as a user`, with nothing on screen explaining that a different principal type was required.

### Steps to Reproduce

1. Create a service account (Admin → Service Accounts), e.g. `svc-raw-ci-write`.
2. Go to the admin **Permissions** page and click **New**.
3. Leave **Principal Type** at its default, **User** (`page.tsx:99` — `EMPTY_FORM` defaults to `"user"`).
4. Open the **Principal** picker and search for the service account — it is listed, labelled with its display name.
5. Select it, pick a repository target, tick `read` + `write`, submit.
6. The form errors with `principal_id … does not exist as a user`.

Workaround: set **Principal Type** to **Service Account** *first*, then re-select the principal — switching the type clears `principal_id` (`page.tsx:422`), so it must be picked again.

### Additional Context

The default principal type being `user` (`page.tsx:99`) makes this the path of least resistance: an operator who doesn't already know service accounts are a distinct principal kind will hit the error before they ever open the type dropdown. The display-name column makes it worse — a service account created with a descriptive `display_name` renders as ordinary prose with no `svc-` prefix visible, since the picker labels by `display_name || username` and only *searches* the username (`page.tsx:171-177`).

Suggested fix, in preference order:

1. Exclude the known service-account ids from the `user` branch of `principalOptions` (`page.tsx:165`) using the `serviceAccountsData` already in scope. Smallest change, no API dependency.
2. Once the backend exposes it (see artifact-keeper#3634), pass `is_service_account=false` to the users query for this picker rather than filtering client-side.
3. Optionally, make the form self-correcting: if the selected principal is a known service account, flip `principal_type` to `service_account` rather than submitting a request that cannot succeed.

For the other four screens consuming `adminApi.listUsers()` — users list, group member picker, audit actor filter, age-gate approver picker — nothing can be done until the API exposes the flag; those are covered by artifact-keeper#3634.

sblaisot:
## Summary

Closes #823.

Service accounts are rows in the `users` table, so `GET /api/v1/users` returns them alongside people, and the admin **Permissions** create form offered them in the principal picker under its default principal type, **User**. That grant can never succeed — the API accepts a service-account row only as `principal_type: "service_account"` and rejects it as a user with `400 principal_id <uuid> does not exist as a user` — and the error arrives after the operator has already picked the target and the actions, naming no remedy.

It bites at the worst possible moment. A CI service account gets `403 You do not have permission to perform this action on this repository` on its first upload (token scopes only narrow; they never grant), and the fix — granting it `write` on the repository — is exactly what this picker blocked.

The page already fetched the service-account list to label existing permission rows, so the ids needed to exclude them were in scope and unused. Changes:

- the **User** branch of `principalOptions` subtracts ids that appear in the service-account list;
- the picker waits for that list before enabling (`"Loading principals..."`), because until it resolves the User branch cannot be filtered and the rejected principal is back on screen for a frame. This also folds the previous service-account-only loading guard into one condition covering both non-group branches.

Selecting **Service Account** as the principal type first was already the workaround. It is no longer needed, and service accounts remain fully available under their own type — that is where the grant legitimately happens.

**Scope and trade-off.** The exclusion is client-side because `GET /api/v1/users` exposes no `is_service_account` field and no filter to ask the server for people only. artifact-keeper#3634 adds both; once an SDK release carries them this becomes a server-side `is_service_account=false` query and stops depending on the page happening to load every user in a single page. The four other admin screens reading the same endpoint — users list, group member picker, audit actor filter, age-gate approver picker — still show service accounts as ordinary users and are **not** fixable here for that same reason; they are covered by artifact-keeper#3634.

## Test Checklist
- [x] Manually tested locally

Three tests in `permissions/page.test.tsx`, under a describe block whose fixtures put the service account in the **users** response — the shape the real API returns:

- *omits them from the User principal list* — **fails on `main`** (`expected <button> to be null`), passes here.
- *does not offer a stale unfiltered list while the accounts load* — **fails on `main`** (`expected false to be true`), passes here. Holds a pending service-accounts promise and asserts the picker stays disabled.
- *still offers them under the Service Account type* — passes both ways by design; it is the guard against over-filtering, so a future change cannot "fix" this by hiding service accounts everywhere.

Verified by reverting `page.tsx` and re-running: 2 failed / 22 passed. With the fix: 24 passed.

`npm run lint` reports 0 errors (35 pre-existing warnings, none in the touched files). `npx tsc --noEmit` reports 22 errors, all pre-existing in the `blast-radius` and `audit` test files and none in the files this PR touches — identical count on a clean `origin/main`.

**Pre-existing failures unrelated to this PR:** the full suite is 8 failed / 3294 passed, in `src/lib/__tests__/cache-time.test.ts` (4), `repo-settings-tab.test.tsx` (3) and `__tests__/repo-detail-downloads.test.tsx` (1). I confirmed the identical 8 failures on a clean `origin/main` checkout, so this PR neither causes nor fixes them.

## UI Changes
- [x] Dark mode verified
- [ ] N/A - no UI changes

No layout, styling or markup structure changes: the picker's option list loses entries and one disabled/label condition is widened, so responsive, dark-mode and a11y behavior are unchanged by construction (the boxes reflect that, not a fresh manual pass in a browser — no dev server was run in this environment). The one user-visible string change is the loading label, `"Loading service accounts..."` → `"Loading principals..."`, since the guard now covers the User branch too.


sblaisot:
Backend counterpart: artifact-keeper#3635 (closes artifact-keeper#3634). It adds `is_service_account` to `AdminUserResponse` plus an `is_service_account` filter on `GET /api/v1/users`, which is what lets this page drop the client-side subtraction for a server-side query — and what unblocks the four other admin screens this PR cannot fix.

brandonrc:
Thanks for the careful diagnosis. Now that backend artifact-keeper#3635 has shipped (1.9.0), the follow-up you described in the PR body is possible: ask the server for people only. I'd like to do that here rather than land the client-side subtraction, for two reasons:
- The Permissions page calls `listUsers()` with no params, so it only ever sees the backend's default 20 rows. Subtracting service accounts from that page shrinks the list further, and users past row 20 are not reachable either way.
- If `GET /service-accounts` fails, the loading guard clears after one retry and the unfiltered list comes back.

Once #826 has merged, here is what I'll push to this branch, keeping your commit, tests and authorship:
1. Request `adminApi.listUsers({ isServiceAccount: false, perPage: 100 })` under its own query key. `["admin-users"]` is shared with the audit page's actor filter, which should keep seeing service accounts.
2. Drop the `serviceAccountIds` subtraction and restore the service-account-only loading guard.
3. Adapt your three tests to the new shape. The "still offers them under the Service Account type" guard stays.
4. Rebase onto current `main` with a single CHANGELOG bullet.

A search-driven picker to lift the 100 cap is a separate issue, which I'll file. I'll squash-merge once CI is green.

Issue #571, opened 2026-07-10
Generic artifact version history UI


brandonrc:
Show/download version history for generic artifacts. Backend: artifact-keeper/artifact-keeper#2367 (`GET /:key/versions/*path`, `?version=<rev|label>` on download/metadata; per-repo versioning_enabled flag).
UI: in the repository artifact view, a versions list (revision, label, size, checksum, uploaded_by, date) with per-version download; a repo setting to toggle versioning_enabled. Extend the existing download-URL/version UI. Part of #567.

Issue #564, opened 2026-07-10
Admin Users List and Group Member Picker Are Effectively Limited to the First Page


danatri:
### Description

The admin users page and the group member add dialog are effectively limited to the first page of users. In instances with more than 20 users, operators cannot reliably reach users at the end of the list because the UI does not page through the full user set or provide a scalable search/filter path.

### Environment Information

- Browser: Chrome
- Browser Version: 149.0.7827.201
- Operating System: windows 11
- Screen Resolution: 1920*1200
- Connection type: (direct, behind proxy) direct

### Web UI Version

1.3.0

### Backend Version

1.2.1

### Expected Behavior

1. The admin users page loads users via `adminApi.listUsers()` without requesting explicit pagination.
2. The group member add dialog reuses the same all-users query for the add-user dropdown.
3. The UI relies on a 20-row table default and does not provide a complete user-browse or add-member flow for larger instances.
4. As a result, users beyond the first visible page are difficult or impossible to select in normal operation.

### Observed Behavior

1. Admin users should be browseable across all pages or via searchable filtering.
2. The group member picker should allow finding and adding any user in the instance.
3. The UI should not depend on a hidden first-page default for core admin workflows.
4. Large instances should remain usable without manual API workarounds.

### Steps to Reproduce

1. Open the admin Users page in an instance with more than 20 users.
2. Confirm the page loads only the initial visible set unless the page is manually refreshed or filtered.
3. Scroll to the bottom of the list and verify that users beyond the first visible page are not reachable through the default flow.
4. Open Admin Groups -> Manage Members for any group.
5. Open the Add Member dropdown and inspect the available users.
6. Try to select a user that is known to exist but is not in the first visible set.
7. Note that the dialog does not provide a scalable search or paging path to reach that user.

### Browser Console Logs

```shell

```

## Impact
1. Admins cannot efficiently manage users after the first visible page.
2. Group membership management becomes unreliable in instances with more than 20 users.
3. Users at the end of the list are effectively unreachable in common workflows.
4. Operators are forced to use API scripts or manual workarounds.

## Evidence in Codebase
- Admin users page calls `adminApi.listUsers()` without explicit pagination parameters.
- Group member dialog reuses the same all-users result for the add-user dropdown.
- The shared table component defaults to 20 rows, which compounds the visibility limit.

## Proposed Fix
1. Make the admin users page explicitly paginated or searchable.
2. Change the group member picker to support search, paging, or server-side lookup.
[...]

PR #826, opened 2026-09-01
fix(users): list people only, not service accounts (#825)

### LINKED ISSUE #825: Admin Users page lists service accounts as people, with person-only actions and an inflated total
### Description

The admin **Users** page (`/users`) lists service accounts as if they were people. They are rows in the `users` table, so `GET /api/v1/users` returns them, and the page renders them with the full set of person-only affordances: **Edit**, **Reset password**, **Force password change**, and the admin/active toggles. None of those are meaningful for a `svc-*` identity — a service account authenticates with API tokens, which are managed on the dedicated **Service Accounts** page (`/service-accounts`), and it is created with `must_change_password: false` and no usable password.

The user count is wrong for the same reason: an instance with 30 people and 12 CI accounts reports 42 users.

This is the second screen affected by the listing defect tracked in artifact-keeper#3634. The first — the Permissions principal picker — was fixed page-locally in #823 / #824 by subtracting known service-account ids client-side. **That technique does not work here**, and that is the point of this issue: the Users page is server-side paginated (#564), so filtering the current page in the browser would leave `total` and `total_pages` counting rows the page no longer shows — an undercounted page and, at the boundary, a phantom trailing page. Correct filtering has to happen in the query that also produces the count.

artifact-keeper#3634 (PR artifact-keeper#3635) adds exactly that: an `is_service_account` filter on `GET /api/v1/users`, applied to both the page query and the pagination count. Once it lands this page can ask the server for people only and keep an accurate total.

### Environment Information

- Browser: Chrome
- Browser Version: current stable
- Operating System: macOS
- Screen Resolution: n/a

### Web UI Version

v1.8.0 (reproduced against `v1.8.0-38-g39ceb05`)

### Backend Version

v1.8.1 / v1.8.2

### Expected Behavior

The Users page lists people. Service accounts are managed on the Service Accounts page, which already offers everything applicable to them (activate/deactivate, rename, delete, token lifecycle). The reported user count reflects people only.

### Observed Behavior

Service accounts appear inline in the Users table with person-only actions available against them, and are included in the total.

### Steps to Reproduce

1. Create a service account (Admin → Service Accounts), e.g. `svc-raw-ci-write`.
2. Open Admin → **Users**.
3. The `svc-*` account is listed alongside real users, with Edit / Reset password / Force password change available, and is counted in the total shown by the pager.

### Additional Context

**Blocked on the backend.** `GET /api/v1/users` has no `is_service_account` filter until artifact-keeper#3635 merges and ships, and the generated SDK (`@artifact-keeper/sdk`, currently `^1.7.0`) will not carry the parameter until it regenerates — `ListUsersData["query"]` today is `search` / `is_active` / `is_admin` / `page` / `per_page`. The web change can be written ahead of the SDK using a narrowed cast (the `versioning_enabled` precedent from #571), and degrades safely: a backend that predates the filter ignores the unknown query parameter and the page behaves exactly as it does today.

**Deliberately not in scope.** The other screens reading `adminApi.listUsers()` should not all be filtered:

- **audit actor filter** — filtering by a service account is legitimate and often the point; CI is frequently the actor under investigation.
- **group member picker** — a group grant is a valid way to give a service account repository access.
- **age-gate approver picker** — arguable, but an approval flow is a human decision; worth its own issue rather than being changed as a side effect here.

**Suggested shape.** Add `isServiceAccount?: boolean` to `ListUsersParams` in `src/lib/api/admin.ts`, forwarded as `is_service_account` only when defined so every other caller's request is unchanged on the wire, and have the Users page pass `false`. Note that `src/lib/api/__tests__/admin.test.ts` asserts the exact `query` object passed to the SDK, and `users/__tests__/page.test.tsx` asserts the exact arguments to `listUsersPage`, so both need updating with the change.

No toggle is proposed: this page has no search box, so nothing can silently return empty, and the Service Accounts page is one click away in the same nav group.

sblaisot:
## Summary

Closes #825. **Do not merge before artifact-keeper#3635** — see *Blocked on* below.

Service accounts are rows in the `users` table, so `GET /api/v1/users` returned them and the admin **Users** page rendered them inline with the full set of person-only actions — **Edit**, **Reset password**, **Force password change**, and the admin/active toggles. None of those mean anything for a `svc-*` identity: it authenticates with API tokens and is created with no usable password, and everything applicable to it already lives on the **Service Accounts** page. The user count was inflated by the same rows — 30 people plus 12 CI accounts reported as 42 users.

The page now asks the server for people only, and the total it shows counts people.

**This is the one that could not be fixed in the browser.** It is why this waited on the backend instead of shipping alongside the Permissions picker fix (#823 / #824), which subtracts known service-account ids client-side. That technique is wrong here: this listing is server-side paginated (#564), so dropping rows from a fetched page would leave `total` and `total_pages` counting rows the page no longer shows — a short page and, at the boundary, a phantom trailing one. The filter has to be applied by the query that also produces the count, which is exactly what artifact-keeper#3634 added (and why that PR filters both statements).

### Blocked on

- **artifact-keeper#3635** must merge and ship. Until a backend carries the filter, this PR is inert — the unknown query parameter is ignored and the page behaves exactly as it does today. So it is *safe* to merge early, it just does nothing. Tagged `needs-backend`.
- `is_service_account` is not in the generated SDK (`@artifact-keeper/sdk@^1.7.0` types `ListUsersData["query"]` as `search` / `is_active` / `is_admin` / `page` / `per_page`), so `adminApi.listUsersPage` builds the query as a plain object and casts once at the boundary — the `versioning_enabled` stopgap from #571. Worth a follow-up to drop the cast when the SDK regenerates.

### Scope

The filter is opt-in at the API layer (`ListUsersParams.isServiceAccount`), sent **only when the caller asks**, so every other call site's request is byte-identical on the wire. Only the Users page passes it. Deliberately unchanged:

- **audit actor filter** — filtering by a service account is legitimate and usually the point; CI is often the actor under investigation.
- **group member picker** — a group grant is a valid way to give a service account repository access.
- **age-gate approver picker** — arguable, but an approval flow is a human decision; that deserves its own issue rather than being changed as a side effect here.

No "show service accounts" toggle: this page has no search box, so nothing can silently return empty, and the Service Accounts page is one click away in the same nav group. Easy to add later if operators miss the combined view.

## Test Checklist
- [ ] Manually tested locally

Two new tests, three updated:

- `admin.test.ts` → *listUsersPage forwards is_service_account only when the caller asks* — asserts the exact query object in both directions, including that the key is **absent** (not `undefined`) when unset, which is what keeps the other four call sites unchanged.
- `users/page.test.tsx` → *asks the server for people only, on every page* — pins that the filter is server-side and that the page does not filter locally.
- Three existing assertions of the exact `listUsersPage` arguments (initial load, next page, page-size change) now include `isServiceAccount: false`. That churn is the point: the filter must apply on every page, not just the first.

**Verified as regression tests:** reverting `admin.ts` and `users/page.tsx` while keeping the tests gives 5 failed / 130 passed; with the fix, 135 passed.

Not manually exercised in a browser — no dev server or live backend in this environment, and the backend filter does not exist yet, so there is nothing to observe end-to-end until artifact-keeper#3635 ships. That box is left unchecked deliberately.

`npm run lint`: 0 errors (35 pre-existing warnings, none in the touched files). `npx tsc --noEmit`: 22 errors, all pre-existing in `blast-radius` and `audit` test files, none in the files this PR touches — identical count on a clean `origin/main`.

**Pre-existing suite failures unrelated to this PR:** 8 failed / 3293 passed, in `src/lib/__tests__/cache-time.test.ts` (4), `repo-settings-tab.test.tsx` (3) and `__tests__/repo-detail-downloads.test.tsx` (1). Confirmed identical on a clean `origin/main` checkout.

## UI Changes
- [x] Dark mode verified
- [ ] N/A - no UI changes

No markup, layout or styling changes — the table renders fewer rows and the pager reports a smaller total, so responsive, dark-mode and a11y behavior are unchanged by construction. The boxes reflect that reasoning, not a fresh manual pass in a browser.


brandonrc:
Thanks. Backend artifact-keeper#3635 merged on 2026-09-02 and shipped in backend 1.9.0 and 1.10.0, so the `is_service_account` filter is live. This PR matches it.

Here is what I'll push to this branch, keeping your commit and authorship:
1. Rebase onto current `main`. The only conflict is `CHANGELOG.md`. The branch's `[Unreleased]` block still holds bullets that were released in 1.9.0 and 1.10.0, so I'll drop the hunk and re-add your single bullet under `[Unreleased] / ### Fixed`.
2. Update the comment on the cast in `admin.ts` to the standard `not in the generated SDK yet (backend 1.9.0, artifact-keeper#3634)`. The cast stays, because `@artifact-keeper/sdk` 1.7.0 still has no `is_service_account`.

I'll squash-merge once CI is green. #824 goes right after, rebuilt on this filter.

PR #838, opened 2026-09-09
test(e2e): refresh the admin monitoring desktop visual baseline

### LINKED ISSUE #835: CI: visual regression 'admin pages › monitoring - desktop' snapshot is stale (918px vs 720px), failing E2E on every open PR
Every open PR since at least #811 fails `E2E Visual Regression` (and therefore the `E2E` aggregate, `success success failure success`) on the same assertion:

```
[visual] › e2e/suites/visual/pages/admin-pages.spec.ts:20:9 › Visual regression: admin pages › monitoring - desktop
Error: expect(page).toHaveScreenshot(expected) failed
Expected an image 1280px by 918px, received 1280px by 720px. 13548 pixels (ratio 0.02 of all image pixels) are different.
```

The last green `CI` run on `main` is from 2026-08-25; the monitoring page's rendered height changed after that (or the baseline was captured at a different viewport), and the committed `monitoring-desktop-admin.png` baseline was never refreshed. #826 and #824 passed because their runs predate the change.

Ask: regenerate the baseline (`npx playwright test --update-snapshots` for that spec on the CI image), or fix the page height regression if it is one, and add the check to the required set only once it is stable. Until then every PR shows a red E2E for a reason unrelated to its diff (e.g. #834).

brandonrc:
Closes #835

## Cause

The committed `monitoring-desktop-admin-visual-linux.png` (1280x918, written by #783 on 2026-08-12) shows the Monitoring page's two empty states ("No alert data" / "No health log entries"). What CI renders today is the populated layout: a `database — OK` alert card, a `1 healthy` badge, and health-log rows, which fits the 720px viewport, hence "Expected 1280x918, received 1280x720".

Nothing in this repo touched the page, the admin components, or the visual suite since the last green run; the page is not missing content, it has more. The backend's scheduler runs the first service health check 15–44s after boot (`jittered_startup_delay(15)`), and the visual screenshot lands ~40–50s after boot (in the failing run the health-log row is timestamped 12:34:46 and the screenshot was taken at 12:34:50). So the empty-state baseline and the populated render were two outcomes of a startup race, and the E2E stack's timing has settled on the populated side. Not a page regression (b): a stale baseline (a) plus a nondeterministic fixture.

## Change

- `e2e/setup/seed-data.ts`: the seed now records one health check via `POST /api/v1/admin/monitoring/check`, so the populated layout is guaranteed before any screenshot regardless of scheduler jitter.
- Regenerated `monitoring-desktop-admin-visual-linux.png` (1280x720) with `npx playwright test e2e/suites/visual/pages/admin-pages.spec.ts -g "monitoring - desktop" --update-snapshots` against the seeded state.

Measured: a render with one more or one fewer health-log row stays within the spec's 1% `maxDiffPixelRatio`; the timestamp text costs a few hundred pixels. The empty-state layout (a size mismatch, which no tolerance can absorb) can no longer occur.

## Verification

Environment matches CI's `e2e-visual` job: Ubuntu 24.04 host, Node 22, `@playwright/test` 1.62.1 with its pinned Chromium (1234), `docker-compose.e2e.yml` with the current `backend:dev` (built 2026-09-08, "Server 1.9.0"), `PLAYWRIGHT_BASE_URL=http://localhost:3100`. Before the change, the full visual project reproduced CI's failure to within 30 pixels (13518 differing pixels locally vs 13548 in CI) and the local render differs from CI's uploaded `monitoring-desktop-admin-actual.png` by 374 pixels (timestamp/version text only).

- `npx playwright test --project=visual` on two fresh stacks: 33 passed, 33 passed.
- The single spec re-run on the same stack after generation (one extra row): passed.
- `npm run lint`: 0 errors (37 pre-existing warnings, none in the changed file).


brandonrc:
## Audit — ready for maintainer review

Root cause is a stale baseline sitting on a race, not a page regression: nothing touching the monitoring page or the visual suite changed since the last green run; the 918 px baseline (from #783) captured the page's two empty states, while CI now usually renders the populated layout (alert card + health-log row, 720 px) because the backend scheduler's first health check lands 15–44 s after boot and the screenshot ~40–50 s after boot. A PNG-only refresh would still flip on high jitter, so the fix also seeds one health check before the visual run. Diff is exactly that: one exported `seedHealthCheck` (`POST /admin/monitoring/check`) called from the seed step, plus the regenerated PNG.

Reproduced in a CI-equivalent stack (Ubuntu 24.04, Node 22, Playwright 1.62.1, `docker-compose.e2e.yml` with `backend:dev`), full visual project 33/33 on two fresh stacks, lint clean. Unblocks the `E2E`/`E2E Visual Regression` reds on every open PR (#811 onward).

PR #840, opened 2026-09-10
feat(repositories): three-state visibility control

### LINKED ISSUE #839: Repository visibility UI: replace the binary public/private switch with three states
## Problem

The repository create dialog, edit dialog and Settings tab all render a binary
"Public repository" switch. It can express only "anyone, including anonymous"
or "grant holders only". It has no way to say "everyone signed in", which is
what most repositories on a corporate instance actually are.

The repository list row has the matching problem: a padlock means "not public",
so a repository readable by every signed-in user renders identically to one
readable by a named few. That is the single most consequential difference a list
row can show.

The backend adds a first-class `visibility` axis (`public` / `internal` /
`private`) — see the backend issue. The UI needs to catch up, and the API
client needs to actually carry the field.

## Scope

- A shared visibility control used verbatim by all three surfaces so they cannot
  drift, each option carrying a one-line hint naming who can actually read the
  repository.
- The list row distinguishes all three states, with an `aria-label` spelling out
  who can read.
- `src/lib/api/repositories.ts` must forward `visibility` on create and update
  and read it back in `adaptRepository`. Both body builders are explicit
  allowlists rather than spreads and `adaptRepository` maps field by field, so an
  unforwarded field is dropped **silently** — the generated SDK types do not
  declare `visibility` yet, so TypeScript will not catch it.
- Read must fall back to deriving the state from `is_public` when the backend
  predates the field, so the UI stays safe to deploy ahead of the server. A
  boolean cannot express `internal`, which is the documented degradation.
- When guest access is disabled, keep the control visible and withdraw only the
  `Public` option — with three states the choice between `Internal` and
  `Private` is still meaningful and still the operator's. A repository already
  `Public` from before the policy changed must keep rendering its own value, or
  opening its settings would silently move it on save.

## Out of scope / follow-ups

The repository detail header and the CVE blast-radius page also derive their
label from `is_public`; those are tracked separately.

mymarche:
Closes #839

## Summary

Replaces the binary "Public repository" switch with a three-state **Visibility**
control (`Public` / `Internal` / `Private`), backing the first-class visibility
axis the backend adds in artifact-keeper#3812 (migration 217).
`Internal` is the state the switch could not express: readable by every
signed-in user with no grant, never readable anonymously.

- New shared `visibility-select.tsx`, used verbatim by the create dialog, the
  edit dialog and the Settings tab so the three surfaces cannot drift — the same
  shared-component pattern as `format-config-fields.tsx`. Each option carries a
  one-line hint naming who can actually read the repository.
- The repository list row distinguishes all three states. A padlock previously
  meant "not public", which would have rendered `internal` identically to
  `private` — the most consequential difference a list row can show. `Public` is
  unmarked as before, `Internal` shows a group icon, `Private` keeps the
  padlock, and both carry an `aria-label` spelling out who can read.
- With guest access disabled the control stays visible and only `Public` is
  withdrawn. The switch used to be hidden entirely and replaced with "Public
  repositories are disabled by the operator", which made sense with two states
  and one unavailable; with three, the choice between `Internal` and `Private`
  is still meaningful and still the operator's. A repository already `Public`
  from before the policy changed keeps rendering its own value, so opening its
  settings cannot silently move it on save.
- The control writes the backend's `visibility` field rather than the deprecated
  `is_public` boolean, and falls back to deriving the state from `is_public`
  when reading a backend that predates the field, so the UI is safe to deploy
  ahead of the server. A boolean cannot express `internal`, which is the
  documented degradation.

### The bug this PR nearly shipped with

`src/lib/api/repositories.ts` is the only path between the UI and the backend,
and it was **not** updated by the first pass. Its create/update bodies are
explicit allowlists rather than spreads, and `adaptRepository` maps field by
field, so `visibility` was dropped in **both** directions — silently, because
the generated SDK types do not declare the field yet, so TypeScript was
satisfied by the local request type that does.

The effect was that the feature did not work at all: create and update sent
*neither* field (the dialogs had stopped sending `is_public`), so choosing a
visibility was a no-op; and `Repository.visibility` was always `undefined` on
read, so every `internal` repository rendered as Private and the new list marker
was unreachable on real data.

Fixed in its own commit, with six unit tests pinning both directions, the
omitted-key case (a legacy `is_public`-only caller must not trip the backend's
contradictory-input rejection) and the unknown-state case. Recording it here
because the shape is worth knowing: a full green unit suite did not catch it —
component tests stop at the submit callback, and the API-level e2e specs never
touch the client. Follow-up 3 below adds the browser-level test that would have.

## Test Checklist
- [x] Unit tests added/updated — `visibility-select.test.tsx` (13 cases, new),
      a `visibility marker` block in `repo-list-item.test.tsx` (7 cases), the
      three-state cases in `repo-dialogs.test.tsx` and
      `repo-settings-tab.test.tsx`, and six in
      `src/lib/api/__tests__/repositories.test.ts` for the wire format.
      Full suite: 177 files / 3312 tests pass. `npm run lint`: 0 errors.
- [x] E2E Playwright tests added/updated — `internal-repo-visibility.spec.ts`
      and `internal-repo-visibility-authenticated.spec.ts`, registered in
      `playwright.config.ts` under `roles-unauthenticated` / `roles-restricted`;
      `e2e-internal-pypi` and `e2e-internal-docker` added to the seed. Also
      fixed `repo-edit.spec.ts`, which asserted the removed switch by matching
      `/public/i` and could no longer pass.
      **These specs have been written and discovered (`--list` finds 20 tests)
      but NEVER EXECUTED** — the suite needs a deployed instance carrying
      migration 217, which was not available. Listed under Follow-ups below.
- [ ] Manually tested locally — not possible: no deployed instance was
      available. Verification is the unit suite plus the API-level specs.
- [x] No regressions in existing tests — 3312/3312 pass. Changed-line coverage
      measured with the `ci.yml` gate's own algorithm: **100% (22/22
      instrumented changed lines)**, above the 80% floor.

### CI status, and why the e2e jobs are red

`Unit Tests`, `Lint`, `Build`, `CodeQL` and the linked-issue gate are green.
Three e2e jobs are not, for three different reasons — only one of which was a
defect in this branch:

1. **`private-repo-visibility.spec.ts`, 5 tests — a regression this branch
   introduced, now fixed.** The seed had replaced `is_public: true` with
   `visibility: 'public'` on `e2e-public-pypi`. The e2e stack runs against
   `artifact-keeper-backend:dev`, built from upstream `main`, which predates the
   visibility column; `CreateRepositoryRequest` does not use
   `deny_unknown_fields`, so it ignored the field and created the repository
   PRIVATE, breaking five pre-existing anonymous-read assertions that have
   nothing to do with this change. The seed now sends both fields — the backend
   rejects the pair only when they contradict — so it is correct against a
   backend on either side of the migration.
2. **`internal-repo-visibility*.spec.ts`, 7 tests — an ordering dependency, not
   a defect.** They assert behaviour that only exists once the backend migration
   ships. They cannot pass until artifact-keeper#3813 merges and a new `:dev`
   image is published. **Merge the backend PR first.** Deliberately NOT
   feature-detected into a skip: a silently-skipping test is the fiction-green
   failure mode this codebase already has scar tissue about (#2924).
3. **3 visual snapshots** (`repositories-mobile-admin`, `dashboard-mobile-admin`,
   `analytics-desktop-admin`) — the seed adds two repositories, which shifts the
   list and the counts on the dashboard and analytics pages. Snapshots are
   platform-specific (`-linux.png`) and cannot be regenerated off CI; they need
   a `--update-snapshots` run after the backend lands.

## UI Changes
- [ ] Playwright E2E spec covers the change — the API-level specs cover the
      backend contract, but **no browser-level spec drives the create → save →
      list-marker path**. That is the gap that let the API-client bug through;
      listed under Follow-ups below.
- [ ] Responsive layout verified (mobile + desktop) — not verified; no deployed
      instance. The control is a standard shadcn `Select` in the same form grid
      as the existing Format/Type selects. The two `/repositories` visual
      snapshots WILL need regenerating: the seed adds two repositories and the
      row marker changed. Listed under Follow-ups below.
- [ ] Dark mode verified — not verified; no deployed instance. No new colours
      are introduced (`text-muted-foreground` throughout).
- [ ] Accessibility checked — partially. Keyboard navigation is the Radix
      `Select` default and the label is tied to the trigger by `id`. One known
      wart: `repo-list-item.tsx` puts `aria-label` on a bare lucide `<svg>` with
      no `role="img"`. `getByLabelText` matches it, so the unit tests pass, but a
      screen reader may announce nothing — this is the only place in the
      codebase that labels a bare SVG. Not fixed here to keep the diff to the
      visibility axis; listed under Follow-ups below.

## Known gaps recorded rather than hidden

- The repository detail header and the CVE blast-radius page still derive their
  label from `is_public`, so both will call an `internal` repository "Private".
  The blast-radius one is security-facing — the backend's classifier now has a
  fourth `internal` state precisely so an operator triaging a CVE is told the
  whole instance can pull the artifact. Listed under Follow-ups below.
- `@artifact-keeper/sdk` has no `visibility` field yet, so this PR carries it
  through defensive casts (the pattern already used for `format_key`). The SDK
  regeneration is blocked on a repository that was not available; tracked in
  follow-up 5 below.
- The `ci.yml` duplication gate does not analyse `.tsx` files at all — run with
  its exact command, jscpd v4 measured only the one changed `.ts` file and
  silently skipped all four components. Pre-existing, same class as backend
  #3495; listed under Follow-ups below.

### Follow-ups (recorded here; not filed as issues)

Deliberately left open, each self-contained:

1. **Regenerate the progenitor SDK.** `xtask` reads the spec from a sibling
   `artifact-keeper-api` repository holding the committed `openapi.json` export,
   which was not available. `RepositoryVisibility` already derives `ToSchema` and
   is registered, so the spec picks the field up as soon as it can be
   regenerated. Until then `@artifact-keeper/sdk` has no `visibility` field and
   the web client carries it through defensive casts.
2. **Run the two internal-visibility e2e specs.** Written, registered in
   `playwright.config.ts`, discovered by `--list` (20 tests), never executed —
   they need a deployed instance carrying migration 217.
3. **Add a browser-level e2e for create -> save -> list marker.** Both new specs
   use only the `request` fixture, so nothing drives the UI. That is precisely
   why the API client dropping `visibility` in both directions survived a full
   green unit suite.
4. **`repo-detail-content.tsx` and the blast-radius page still read
   `is_public`,** so both label an `internal` repository "Private". The
   blast-radius one is security-facing: the backend classifier now has a fourth
   `internal` state exactly so a CVE triage is not told the artifact reached a
   handful of grantees. `src/types/search.ts` and `tree.ts` likewise carry only
   the boolean.
5. **The web duplication gate does not analyse `.tsx` at all.** With ci.yml's
   own command, jscpd v4 measured only the one changed `.ts` file and silently
   skipped all four components; `--format tsx`/`jsx` produce no report. Same
   class as backend #3495, which was fixed there by failing when a changed file
   is missing from the report.
6. **`proxy_helpers.rs` still gates the virtual-member read on
   `member.is_public`,** so an `internal` member is invisible through a virtual
   parent to a caller who can read it directly. Fails closed, so it is a
   uniform-enforcement bug, not a hole. Same shape in `sbom.rs` and `conda.rs`.
   Note for whoever picks this up: the tenant/config gates in `repo_tokens.rs`,
   `promotion.rs` and `curation.rs` deliberately still read `is_public` and must
   NOT be converted — `internal` has to behave as `private` there.
7. **Port `docs/repository-visibility.md` to the public docs site,** which is
   built from a separate repository.


brandonrc:
Thanks for this, and for listing the known gaps so clearly. The wire contract matches backend artifact-keeper#3813 at its current head (f2a0ac89): field name, values, request shapes, the 400 on contradictory input, and the response field.

The backend PR is still in review, so this one stays open until it merges and a new `:dev` image is published. Meanwhile, here is what I'll push to this branch, keeping your commits and authorship:
1. **Send `is_public` alongside `visibility`** (`is_public: visibility === "public"`). Backend main ignores unknown keys. Today, switching Public→Private on a backend without #3813 does nothing, yet the UI still says "Repository updated". With both fields, the change takes effect, and `internal` degrades to private.
2. **Send `visibility` from the edit dialog only when it changed.** Otherwise, renaming an already-public repo with guest access disabled resends `public`, and the backend coerces it to private.
3. **Show three states on the blast-radius page and the repo detail header.** Use `access_scope` for the Visibility column, add an `internal` badge, and word the "everyone" alert for internal repos. Both pages call an internal repository "Private" today. The blast-radius one is security-facing, so it needs to ship with the control that makes internal repos possible.
4. **Update comments and messages.** The backend now coerces a disabled-guest Public request to `private`, not `internal`. I'll also reference artifact-keeper#3813 instead of "migration 217", since that number is taken on main and the backend review renumbers it.
5. **Add `role="img"`** to the two labelled list-row icons.
6. **Rebase onto current `main`** with a single CHANGELOG bullet `(#839, backend artifact-keeper#3813)`.
7. **After the backend lands:** re-run E2E and regenerate the three visual snapshots.

The `.tsx` gap in the duplication gate is real and predates this PR. I'll file it separately.

brandonrc:
Correction to my note above about the duplication gate: I checked, and it does analyse `.tsx`. On #794, CI's jscpd step reported "0.0% (0/1024 duplicated lines)" over one `.ts` file and `repo-setup-guide.tsx`, which is about 1,000 lines on its own. Locally, jscpd 4.3.0 with CI's flags finds the same clones in `.tsx` as in `.ts`. So I will not file an issue for it. The earlier observation was probably an older jscpd 4.x.

PR #852, opened 2026-09-18
fix: lazy-load artifact tree and preserve search

### LINKED ISSUE #850: Bug: Artifact tree view only shows first 100 files in a folder
When a generic/local repository contains more than 100 files in the same directory, the Artifact Keeper tree view only shows 100 files.
The remaining files exist in the repository and were uploaded successfully, but they are not visible in the tree view.
This also happens when the directory is initially collapsed and then expanded.

Environment
    Artifact Keeper version backend: 1.9.1
    Artifact Keeper version frontend: 1.9.0
    Repository format: generic
    Repository type: local
    Repository key: test
    Browser: Brave 1.95.101 
    Deployment: Docker Compose

Steps to reproduce

1. Create a generic/local repository named test.

2. Generate 1000 files:

```
mkdir -p test-files
for i in $(seq -w 1 1000); do
    printf 'test file %s\n' "$i" > "test-files/file-${i}.bin"
done
```

3. Upload all files into the same directory:

```
BASE_URL="https://artifacts.example.com"
REPOSITORY="test"
TOKEN="<token>"

for file in test-files/*; do
    filename=$(basename "$file")

    curl \
        --fail-with-body \
        --silent \
        --show-error \
        -H "Authorization: Bearer ${TOKEN}" \
        -T "$file" \
        "${BASE_URL}/api/v1/repositories/${REPOSITORY}/artifacts/loadtest/${filename}"
done
```

4. Expected behavior

All 1000 files should eventually be accessible from the tree view.

Doubleth1nk:
## Summary
Fixes the repository tree view for repositories containing more artifacts than the artifact-list pagination limit.

Previously, the tree view was constructed client-side from the regular artifact listing. The frontend requested a large page of artifacts, but the backend limits artifact-list responses to 100 items. As a result, repositories with more than 100 artifacts could show an incomplete folder tree.

This change switches the repository tree view to the dedicated /api/v1/tree endpoint and loads folder contents on demand.

This fixes #850 

## Test Checklist
- [X] Manually tested locally


brandonrc:
Thanks for this. Switching the tree to `/api/v1/tree` is the right fix for #850, and I checked the adapter against the backend handler. The endpoint returns every child with no 100-item cap. File nodes come back as `type: "file"`, and your adapter maps them correctly.

Here is what I'll push to this branch as a maintainer, keeping your commits and authorship:
1. Rebase onto current `main`. There is one conflict in `repo-detail-content.tsx`, where main added the package-analysis block next to your page-size change.
2. Add a CHANGELOG bullet under `[Unreleased] / ### Fixed`.
3. Add a `getChildren` test where the backend returns `type: "file"`. The current fixtures send `artifact`, which the backend never does.
4. Delete `src/lib/artifact-tree.ts` and its test, since nothing imports them any more.
5. Add a test for searching while in Tree view, so the changed lines stay clearly above the 80% coverage gate.

The red `E2E Interactions (shard 3/3)` comes from `age-policy.spec.ts`, which main has since fixed. It is unrelated to this change. I'll squash-merge once CI is green.

